Issue #2, opened 2022-09-23
API feedback


srid:
I tried using this as a [flake-parts module](https://github.com/numtide/treefmt-nix/issues/1). Relevant PRs,

- https://github.com/srid/treefmt-flake/pull/6 (flake-parts module)
- https://github.com/srid/haskell-template/pull/58 (consumer of the above module)

## Why is `build` an option?

In the first PR, you'll notice the following code:

```nix
              let mod = inputs.treefmt-nix.lib.evalModule pkgs { projectRootFile = "flake.nix"; };
              in builtins.removeAttrs mod.options [
                "_module"
                # TODO: Why is this an option if the user can't set it?
                "build"
              ];
```

This is the only way I could get access to the combined `options` (to pass to `flake-parts` module as is), but I was forced to hack around by removing the `_module` attribute, as well as `build` option. If I leave the `build` option around, the consumer of my flake-parts module is asked to explicitly set `build.wrapper` and friends (which doesn't make sense of course).

**Why does the `build` option even exist if the user is not supposed to set them?**

Also note that I'm forced to pass `projectRootFile = "flake.nix";`, so I'd think **we may want to export the "final" options set in `lib`** so consumers of this flake do not have to do all of the above. I'm not yet too familiar with the module system to come up with a cleaner approach as first attempt.

## Accumulate formatter packages

This piece of snippet in the second PR is used to accumulate all formatter packages, along with the wrapper itself so that they can be put in the dev shell. This enables non-treefmt-aware editors to directly invoke the formatters. I imagine this can be **upstreamed somehow here**?

```nix
              treefmtConfig = (inputs.treefmt-nix.lib.evalModule pkgs config.treefmt).config;
              treefmtTools = {
                treefmt = treefmtConfig.build.wrapper;
              } // builtins.mapAttrs (_: v: v.package) treefmtConfig.programs;
```

## `programs` and `settings` duality is unnecessary and problematic

I don't understand why this module provides **two** top-levels for configuring treefmt -- `programs` and `settings`. They even have overlapping options, ie.: `programs.foo.package` and `programs.foo.command`. I find that it is impossible to use **both**. For example, if you comment the multiline comment in the second PR it will lead to:

```
error: The option `perSystem.aarch64-darwin.treefmt.settings.formatter.ormolu.command' is used but not defined.
(use '--show-trace' to show detailed location information)
```

[...]

PR #14, opened 2022-12-21
Add `flake-parts` module

### LINKED ISSUE #1: Add a flake-parts module?
Is it appropriate to upstream https://github.com/srid/treefmt-flake to this repo? It mainly adds the flake check for testing that the project tree is already autoformatted.

srid:
I'm opening a draft PR (instead of waiting for full implementation) just to get something working, as well as to get any early feedback. The commits in this PR can be squash merged (or I can rebase them if necessary) once ready.

This PR adds a `flake-parts` module for working with treefmt in the most ergonomic way. It also adds a flake check to test in CI that the source tree is formatted. You can see a demo here: https://github.com/srid/haskell-template/pull/75

Resolves #1 

- [x] Can the `types.raw` be replaced with appropriate types? See also: https://github.com/numtide/treefmt-nix/issues/2
- [x] Add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ This flake exposes a [flake-parts](https://flake.parts/) module as well. To use 
 
 1. Add `inputs.treefmt-nix.flakeModule` to the "imports" list of your `flake-parts` call.
 1. Add `treefmt = { .. }` (containing the configuration above) to your "perSystem" section.
-1. Add `config.treefmt.build.wrapper` to the `buildInputs` of your devShell. This will make the `treefmt` command available in the shell using the specified configuration.
+1. Add `config.treefmt.build.wrapper` to the `nativeBuildInputs` of your devShell. This will make the `treefmt` command available in the shell using the specified configuration.
     - You can also use `config.treefmt.build.programs` to get access to the individual programs, which could be useful to provide them to your IDE or editor.
 
 For an example, see [haskell-template](https://github.com/srid/haskell-template)'s `flake.nix`.
--- a/README.md
+++ b/README.md
@@ -25,6 +25,17 @@ treefmt-nix.lib.mkWrapper nixpkgs {
 }
 ```
 
+### `flake-parts`
+
+This flake exposes a [flake-parts](https://flake.parts/) module as well. To use it:
+
+1. Add `inputs.treefmt-nix.flakeModule` to the "imports" list of your `flake-parts` call.
+1. Add `treefmt = { .. }` (containing the configuration above) to your "perSystem" section.
+1. Add `config.treefmt.build.wrapper` to the `buildInputs` of your devShell. This will make the `treefmt` command available in the shell using the specified configuration.
+    - You can also use `config.treefmt.build.programs` to get access to the individual programs, which could be useful to provide them to your IDE or editor.
+
+For an example, see [haskell-template](https://github.com/srid/haskell-template)'s `flake.nix`.
+
 ## Supported programs
 
 <!-- `> ls programs/*.nix | grep -v default.nix | cut -d '.' -f 1 | cut -d / -f 2 | sort | sed -e 's/^/* /'` -->
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -1,4 +1,3 @@
-{ treefmtLib }:
 { self, lib, flake-parts-lib, ... }:
 let
   inherit (flake-parts-lib)
@@ -8,7 +7,6 @@ let
     types;
 in
 {
-  _file = __curPos.file;
   options = {
     perSystem = mkPerSystemOption
       ({ config, self', inputs', pkgs, system, ... }: {
@@ -26,7 +24,7 @@ in
                 type = types.raw; # TODO: module type?
                 description = "Evaluated module of treefmt-nix";
                 default =
-                  treefmtLib.evalModule pkgs config.treefmt.config;
+                  (import ./.).evalModule pkgs config.treefmt.config;
               };
               programs = mkOption {
                 type = types.attrsOf types.package;
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -53,7 +53,7 @@ in
     perSystem = { config, self', inputs', pkgs, ... }: {
       checks.treefmt = pkgs.runCommandLocal "treefmt-check"
         {
-          buildInputs = [ pkgs.git config.treefmt.wrapper ] ++ config.treefmt.programs;
+          buildInputs = [ pkgs.git config.treefmt.wrapper ] ++ lib.attrValues config.treefmt.programs;
         }
         ''
           set -e
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -52,19 +52,14 @@ in
             }
             ''
               set -e
-              # treefmt uses a cache at $HOME. But we can use --no-cache
-              # to make treefmt not use a cache. We still seem to need
-              # to export a writable $HOME though.
-              # TODO: https://github.com/numtide/treefmt/pull/174 fixes this issue
-              # but we need to wait until a release is made and that release gets
-              # into the nixpkgs we use.
-              export HOME="$TMP"
+              treefmt --version
               # `treefmt --fail-on-change` is broken for purs-tidy; So we must rely
               # on git to detect changes. An unintended advantage of this approach
               # is that when the check fails, it will print a helpful diff at the end.
-              cp -r ${self} $HOME/project
-              chmod -R a+w $HOME/project
-              cd $HOME/project
+              PRJ=$TMP/project
+              cp -r ${self} $PRJ
+              chmod -R a+w $PRJ
+              cd $PRJ
               git init
               git config user.email "nix@localhost"
               git config user.name Nix
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -47,39 +47,37 @@ in
             };
           };
         };
+        config = {
+          checks.treefmt = pkgs.runCommandLocal "treefmt-check"
+            {
+              buildInputs = [ pkgs.git config.treefmt.wrapper ] ++ lib.attrValues config.treefmt.programs;
+            }
+            ''
+              set -e
+              # treefmt uses a cache at $HOME. But we can use --no-cache
+              # to make treefmt not use a cache. We still seem to need
+              # to export a writable $HOME though.
+              # TODO: https://github.com/numtide/treefmt/pull/174 fixes this issue
+              # but we need to wait until a release is made and that release gets
+              # into the nixpkgs we use.
+              export HOME="$TMP"
+              # `treefmt --fail-on-change` is broken for purs-tidy; So we must rely
+              # on git to detect changes. An unintended advantage of this approach
+              # is that when the check fails, it will print a helpful diff at the end.
+              cp -r ${self} $HOME/project
+              chmod -R a+w $HOME/project
+              cd $HOME/project
+              git init
+              git config user.email "nix@localhost"
+              git config user.name Nix
+              git add .
+              git commit -m init
+              treefmt --no-cache
+              git status
+              git --no-pager diff --exit-code
+              touch $out
+            '';
+        };
       });
-  };
-  config = {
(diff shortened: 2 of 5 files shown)


flake-module.nix
```diff
@@ -0,0 +1,85 @@
...
+                description = "Wrapper treefmt using the current configuration";
+                default = config.treefmt.module.config.build.wrapper;
+              };
+            };
+          };
+        };
+      });
+  };
+  config = {
+    perSystem = { config, self', inputs', pkgs, ... }: {
```
roberth, line 53: Everything happens in the `perSystem` module, so you could simplify this to more or less:

```suggestion
  config = {
```

It's a bit of a weird (and incomplete) diff, but the idea is to merge both submodules.

roberth (commented):
Maybe I should read more of the treefmt implementation, but I'm confused by the supposed submodule-ish options in this PR. Doesn't treefmt already do its things in modules, that just need to be imported?

srid:
> Maybe I should read more of the treefmt implementation, but I'm confused by the supposed submodule-ish options in this PR. Doesn't treefmt already do its things in modules, that just need to be imported?

The module options are spread across several files. They are passed to `evalModules` here: https://github.com/numtide/treefmt-nix/blob/4230552f8a71c510d222f1804e42184e106aafd7/default.nix#L23-L25

Is there a straightforward way to "merge" them to produce a single submodule for use in `flake-parts` options?

roberth:
> Is there a straightforward way to "merge" them to produce a single submodule for use in `flake-parts` options?

`imports = [ ]`, or `submoduleWith { modules = [ ... ]; }` instead of `submodule`.

Another trick is you could use `(<<treefmt>>.evalModule foo {}).type`, which is a submodule type with all treefmt stuff in it. This doesn't show intent like it should, so perhaps it's better to refactor `lib.nix` a bit to extract the logic leading up to `lib.evalModules` into a function. Then you can do `submoduleWith { modules = <<treefmt>>.prepareModule foo bar; }`. It's the same thing, but less likely to be accidentally broken.

flake-module.nix
```diff
@@ -0,0 +1,81 @@
...
+              buildInputs = [ pkgs.git config.treefmt.wrapper ] ++ lib.attrValues config.treefmt.programs;
+            }
+            ''
+              set -e
+              # treefmt uses a cache at $HOME. But we can use --no-cache
+              # to make treefmt not use a cache. We still seem to need
+              # to export a writable $HOME though.
+              # TODO: https://github.com/numtide/treefmt/pull/174 fixes this issue
+              # but we need to wait until a release is made and that release gets
+              # into the nixpkgs we use.
```
zimbatm, line 60: AFAIK this is ok now isn't it?

flake-module.nix
```diff
@@ -0,0 +1,81 @@
...
+in
+{
+  options = {
+    perSystem = mkPerSystemOption
+      ({ config, self', inputs', pkgs, system, ... }: {
+        options.treefmt = mkOption {
+          description = ''
+            treefmt flake module options
+          '';
+          type = types.submodule {
```
zimbatm, line 17: do you think it's possible to graft the other module directly into here, so the options can be exposed directly?

zimbatm (commented):
nice

srid:
@roberth Is `flake-parts` supposed to work with NixOS modules, in theory? Because this approach:

> `imports = [ ]`, or `submoduleWith { modules = [ ... ]; }` instead of `submodule`.

gives me:

```
error: The option `build.wrapper' is defined multiple times.
```

---

I'll try the other approach, but now I remember that I have already tried before something like that in vain: https://github.com/Platonic-Systems/treefmt-flake/pull/6/files#diff-0549e7a291dba4b0c8148bb5c499a14bb7c1cd0aa61e74e31623467004ad2b6fR23

srid:
Made some progress:

```nix
              config = mkOption {
                description = "treefmt-nix configuration";
                # type = types.raw; # TODO: Should use module-options.nix
                type =
                  let
                    mod = lib.evalModules {
                      modules = (import ./.).all-modules pkgs;
                    };
                  in
                  types.submodule {
                    options = builtins.removeAttrs mod.options [
                      "_module"
                    ];
                    config = mod.config;
                  };
              };
```

But this now throws:

```
error: The option `perSystem.aarch64-darwin.treefmt.config.programs.cabal-fmt.enable' has conflicting definition values:
       - In `/nix/store/d7538jxb7zw3hawcrwp67g9rvqdgcmp5-source/flake-module.nix': false
       - In `<unknown-file>, via option perSystem': true
```

Yes, at the *use site* `cabal-fmt.enable` is set to `true` ... but somehow that triggres a conflict with `flake-parts` only.

srid:
```
error: The option `build.wrapper' is defined multiple times.
```

Actually, I found a solution to this: need to move the `wrapper` definition from "config" to under "options" (as `default`). Why is that? 🤷🏿 I wish Nix had Haskell-like type system ...

flake-module.nix
srid: @zimbatm I've done this now. `config`'s type is now derived from module-options.nix. 

What do you think of the top-level of `wrapper` option? It just a convenient (shorter-to-type) alias to `module.config.build.wrapper` now. As for the others:

- `programs`: We need this to gather all formatters in one place, so they can be put in the devShell's nativeBuildInputs which is needed to get editors and IDEs (running nix environment) recognize the individual formatters.
- `module`: This is not meant to be used by the user generally, it simply allows us to evaluate the module once rather than twice. Though it could be useful in some advanced cases.

flake-module.nix
```diff
@@ -0,0 +1,78 @@
...
+              wrapper = mkOption {
+                type = types.package;
+                description = "Wrapper treefmt using the current configuration";
+                default = config.treefmt.module.config.build.wrapper;
+              };
+            };
+          };
+        };
+        config = {
+          checks.treefmt = pkgs.runCommandLocal "treefmt-check"
```
zimbatm, line 51: the script could be moved to the main module build.check so it can be re-used outside of flake-parts

flake-module.nix
```diff
@@ -0,0 +1,78 @@
...
+            {
+              buildInputs = [ pkgs.git config.treefmt.wrapper ] ++ lib.attrValues config.treefmt.programs;
```
zimbatm, line 53: Are you sure the programs are needed? I would assume that the treefmt config would already have a store reference to them, which is then referred to by the wrapper.

flake-module.nix
```diff
@@ -0,0 +1,78 @@
...
+                };
+              };
+              module = mkOption {
+                type = types.raw; # TODO: module type?
+                description = "Evaluated module of treefmt-nix";
+                default =
+                  (import ./.).evalModule pkgs config.treefmt.config;
+              };
+              programs = mkOption {
+                type = types.attrsOf types.package;
```
zimbatm, line 32:
```suggestion
                type = types.listOf types.package;
```
A set of all the packages isn't it?

flake-module.nix
zimbatm: if `programs` is move to the main module I think that's enough, then we can get rid of the evalModule call below.

flake-module.nix
srid: True, oversight on my part (the old module [needed it](https://github.com/Platonic-Systems/treefmt-flake/blob/master/flake-module.nix#L48) because it had no wrapper). Fixed.

flake-module.nix
srid: I used an attrset rather than a list - because, while the likes of `buildInputs` does take a list, in other places attrsets are used. 

For example, `haskell-flake`'s `buildTools` is an attrset of packages, because it allows the user to override the default build tools to disable some of them (by setting `{ foo = null; }`. 

```nix
          buildTools = hp: {
            treefmt = config.treefmt.wrapper;
          } // config.treefmt.programs;
```

https://github.com/srid/haskell-template/blob/1706660e7b02ece881a67c99692b70db5ae5c8db/flake.nix#L28-L30

---

I also figured it would be nice to "tag" each program with the formatter it is associated with it, so the caller can do any special processing that way. If we make this use `listOf` instead, then haskell-template will have to re-implement the logic.

srid:
This is ready for review! 

@zimbatm Let me know what decide about the type of the `programs` option.

README.md
```diff
@@ -25,6 +25,17 @@ treefmt-nix.lib.mkWrapper nixpkgs {
 }
 ```
 
+### `flake-parts`
+
+This flake exposes a [flake-parts](https://flake.parts/) module as well. To use it:
+
+1. Add `inputs.treefmt-nix.flakeModule` to the "imports" list of your `flake-parts` call.
+1. Add `treefmt = { .. }` (containing the configuration above) to your "perSystem" section.
+1. Add `config.treefmt.build.wrapper` to the `buildInputs` of your devShell. This will make the `treefmt` command available in the shell using the specified configuration.
```
roberth, line 34:
```suggestion
1. Add `config.treefmt.build.wrapper` to the `nativeBuildInputs` of your devShell. This will make the `treefmt` command available in the shell using the specified configuration.
```

module-options.nix
```diff
@@ -86,32 +86,81 @@ in
           The treefmt package, wrapped with the config file.
         '';
         type = types.package;
+        default =
```
roberth, line 89:
```suggestion
        defaultText = lib.literalMD "wrapped `treefmt` command";
        default =
```

roberth: or leave it in `config`

srid: I'm really curious about the purpose of `defaultText` here. Why add `defaultText` to `wrapper` but not `programs` even though both of them define a `default` value?

roberth: The documentation can't display any packages. So this is to get the docs to build. I should've made that clear.
Feel free to add more defaultTexts, if you think it improves the docs.
If you don't think a default is relevant, you can move the definition to the `config` section, possibly with `lib.mkDefault` if you want to invite users to override it.
Alternatively, you can make options `internal = true` to remove them from the docs.

flake-module.nix
```diff
@@ -0,0 +1,26 @@
...
+          type = types.submoduleWith {
+            modules = (import ./.).all-modules pkgs;
+          };
+        };
+        config = {
+          checks.treefmt = config.treefmt.build.check self;
+        };
+      });
+  };
+}
```
zimbatm, line 26: I like how smaller this module has become. Nice :)

Maybe a last question to @roberth mostly; given that flake-parts has a global namespace, do you think we should prefix treefmt into something like `programs.treefmt` for example?

roberth: I don't mind a flat namespace at all. And maybe treefmt is more than a program?
In Nixpkgs we're actually trying to get rid of the categorization hierarchy for various reasons. RFC coming soon, but preparations are happening here https://github.com/nixpkgs-architecture/simple-package-paths

roberth:
Made a PR to add it to [flake.parts](https://flake.parts) if that's ok.
Not sure about the naming. The repo is `treefmt-nix` but the tool and option names are `treefmt`. What do you think?
 - https://github.com/hercules-ci/flake.parts-website/pull/65
 
I'll add a link to the rendered docs there.

Mic92:
How can I access treefmt itself for interactive usage?

Mic92:
```
perSystem { config }: = {
     packages.treefmt = config.treefmt.build.wrapper;
};
```

Would it make sense to add this output?

Mic92:
https://github.com/numtide/treefmt-nix/pull/18

srid:
@Mic92 Does https://flake.parts/debug.html help? 

I feel like unconditionally adding `packages.treefmt` (#18) is unnecessary in most cases. It can also feel intrusive; for example, I don't want `packages.treefmt` showing up CI jobs of every project of mine using this flake module. Perhaps it can be made conditional based on an enable option?

Mic92:
I think instead of a package it could also become module documentation (i.e. an example that adds the wrapped treefmt to a devshell) but just now it is not very visible without digging into the code how one can access treefmt otherwise.

PR #21, opened 2022-12-27
compress the settings

zimbatm:
Compress the programs and formatters together. This reduces quite a bit of complexity in the formatter folder and also makes formatter settings more directly available.

The main downside is that it's not possible to extend the programs schema with custom settings anymore (see rustfmt and ormolu). The other downsides are that the settings are not cleanly separated and programs
not cleanly exposed anymore (ie: no defaultText).


zimbatm:
What do you think @srid, is that what you had in mind?

Do you have ideas on how to mitigate some of the downsides?

PR #23, opened 2022-12-30
flake-module: add treefmt app output

marijanp:
For convenience I want to have a treefmt app. i.e. I don't want to add treefmt.wrapper to all my devshells and having a app I can simply run it by using `nix run`

EDIT: seems like `nix run` also picks up the `packages` output. IIRC this is not the case for all nix versions, so either close this PR or merge

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -19,6 +19,10 @@ in
           };
         };
         config = {
+          apps.treefmt = {
+            type = "app";
+            program = "${self'.packages.treefmt}/bin/treefmt";
+          };
           checks.treefmt = config.treefmt.build.check self;
           packages.treefmt = config.treefmt.build.wrapper;
         };


flake-module.nix
```diff
@@ -19,6 +19,10 @@ in
           };
         };
         config = {
+          apps.treefmt = {
+            type = "app";
+            program = "${self'.packages.treefmt}/bin/treefmt";
```
srid, line 24: `lib.getExe`?

srid (commented):
> EDIT: seems like `nix run` also picks up the `packages` output. IIRC this is not the case for all nix versions, so either close this PR or merge

I'm not sure what you mean here exactly. But the `packages` output you are talking about here presumably refers to the one added here: https://github.com/numtide/treefmt-nix/pull/18 - which PR I think is not super-useful to most users (I explained why [here](https://github.com/numtide/treefmt-nix/pull/14#issuecomment-1365944045)); treefmt-nix should not be adding a `packages.treefmt` IMO (because the user has not asked for it).

You can instead do:

```nix
apps.treefmt = {
  type = "app";
  program = lib.getExe config.treefmt.build.wrapper;
};
```

However, IMO - this flake module should add neither `packages` and `apps`. Instead, users can do it themselves using `config.treefmt.build.wrapper` (or `.programs`). I tend towards a "Explicit is better than implicit" philosophy here.

zimbatm:
thanks for your interest @marijanp! As srid noted, this is not needed anymore. In order to keep this project lean I will close the PR and let users upgrade their nix installations instead.

PR #31, opened 2023-01-22
flake-module.nix: Fix doc

roberth:
This should make the docs render again.
`lib.mdDoc` isn't needed here. Flake-parts skips the docbook migration, so everything is already markdown.


zimbatm:
thanks robert!

PR #33, opened 2023-01-23
feat: add programs.hlint

teto:
style checker for haskell

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ For an example, see [haskell-template](https://github.com/srid/haskell-template)
 * elm-format
 * gofmt
 * gofumpt
+* hlint
 * nixpkgs-fmt
 * ormolu
 * prettier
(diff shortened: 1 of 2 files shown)


zimbatm:
thanks!

srid:
@zimbatm This PR is interesting, because it deviates from the autoformating domain. hlint is not a formatter; more like a code linter. Is that in scope of treefmt? If so, I might open a PR to add `haskell-language-server` check (which checks that IDE support works) as well. However, this will require a slight change to the `check` output, to make it use the devshell's nativeBuildInputs, because HLS requires Haskell build tools ([cf](https://github.com/srid/haskell-flake/blob/8688fb2deaa193795467eedb4d2473dbeaba19b4/flake-module.nix#L142-L148)).

zimbatm:
If it works as intended my practical side says that it's fine. We started using linters on some projects and it works surprisingly well. But it would be good to clarify linter usage upstream. In the docs, and maybe come up with a different name than "formatter" for the keys then. Let me know if you have some ideas in that area.

PR #41, opened 2023-02-06
[ocamlformat]: Version detection

r-muhairi:
Note
---
There is no way to pass in config file, ``--config`` takes in manually placed options, not ``.ocamlformat`` files.

### DIFF
--- a/programs/ocamlformat.nix
+++ b/programs/ocamlformat.nix
@@ -4,49 +4,46 @@
 , ...
 }:
 let
-  inherit (lib) types;
-
+  l = lib // builtins;
   cfg = config.programs.ocamlformat;
 
   detectVersion = configFile: pkgSet:
     let
       optionValue = list:
-        assert lib.assertMsg (list != [ ]) "treefmt-nix: Unable to detect ocamlformat version from file";
-        lib.elemAt list (lib.length list - 1);
+        assert l.assertMsg (list != [ ]) "treefmt-nix: Unable to detect ocamlformat version from file";
+        l.elemAt list (l.length list - 1);
     in
-    builtins.getAttr "ocamlformat_${
-      builtins.replaceStrings ["."] ["_"]
-      (optionValue (lib.findFirst (option: builtins.head option == "version") []
-          (builtins.map (n: lib.splitString "=" n) (lib.splitString "\n" (builtins.readFile configFile)))))
+    l.getAttr "ocamlformat_${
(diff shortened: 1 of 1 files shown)


programs/ocamlformat.nix
```diff
@@ -4,17 +4,49 @@
...
+  detectVersion = configFile: pkgSet:
+    let
+      optionValue = list:
+        assert lib.assertMsg (list != [ ]) "treefmt-nix: Unable to detect ocamlformat version from file";
+        lib.elemAt list (lib.length list - 1);
+    in
+    builtins.getAttr "ocamlformat_${
+      builtins.replaceStrings ["."] ["_"]
+      (optionValue (lib.findFirst (option: builtins.head option == "version") []
+          (builtins.map (n: lib.splitString "=" n) (lib.splitString "\n" (builtins.readFile configFile)))))
```
zimbatm, line 20: It's a bit problematic that this is depending on an Import From Derivation to work. The `builtins.readFile` is reading the build result of the ocamlformat package that's passed in.

A better solution would be to set the `meta.mainProgram` of the package in nixpkgs. This would make ocamlformat work out of the box.

zimbatm: But thanks for the contribution!

r-muhairi: > It's a bit problematic that this is depending on an Import From Derivation to work. The `builtins.readFile` is reading the build result of the ocamlformat package that's passed in.

I don't follow? builtins.readFile is not reading the build result of the ocamformat package nor is it reading any other derivation, what it is reading however is a path passed as a module option, which should be fine because its known pre-eval.

There is no IFD?

zimbatm: My bad. Can you show me a usage example?

r-muhairi: > My bad. Can you show me a usage example?
```nix
programs.ocamlformat = {
     enable = true;
     package = ./.ocamlformat;
 };
```

zimbatm: I see. Let's split it out into three different options so they can be better documented:

* package - leave that as before. Is only used if the ocamlformat file is not passed.
* pkgs - the package set for ocamlformat file to look into. defaults to nixpkgs.
* ocamlformat (name to be discussed) - type: none or path - path to the .ocamlformat file. Used to pick the right version of ocamlformat if passed.

zimbatm (approved):
thanks!

bors merge

PR #44, opened 2023-02-10
[ocamlformat]: trim option/value

r-muhairi:
My bad, forgot to trim the values to support spaced cases

e.g
```ocamlformat
version =  0.24.1
```


zimbatm (approved):
bors merge

PR #45, opened 2023-02-11
Fix and check docs

roberth:
This makes sure that the docs can be rendered.

The error was as follows, excluding the stack trace:

```
error: Most nixpkgs attributes are not supported when generating documentation.
       Please check with `--show-trace` to see which option leads to this `pkgs.AAAAAASomeThingsFailToEvaluate` reference. Often it can be cut short with a `defaultText` argument to `lib.mkOption`, or by escaping an option `example` using `lib.literalExpression`.
```


zimbatm (approved):
thanks rob!

bors merge

PR #64, opened 2023-04-26
flake-parts: Add `flakeCheck` option to disable adding check

### LINKED ISSUE #63: [flake-parts] Allow disabling the autowiring of flake check
https://github.com/numtide/treefmt-nix/blob/b25d1a3c2c7554d0462ab1dfddf2f13128638b90/flake-module.nix#L36

When https://github.com/cachix/pre-commit-hooks.nix already adds a check (it includes a treefmt hook), the check added by treefmt is redundant and will just slow CI down unnecessarily.

An option in the flake-parts module to disable this check would be nice.

srid:
Resolves #63

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -29,11 +29,18 @@ in
                   Enables `treefmt` the default formatter used by the `nix fmt` command
                 '';
               };
+              options.flakeCheck = lib.mkOption {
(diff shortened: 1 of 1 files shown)


zimbatm (approved):
sweet, thanks!

bors merge

PR #66, opened 2023-05-01
Fix docs

roberth:


zimbatm (approved):
thanks, I keep making that same mistake

PR #73, opened 2023-05-15
Add support for dprint

### LINKED ISSUE #51: Add dprint formatter
I use [dprint](https://dprint.dev/) for TS/JS/JSON/Markdown. It’s really fast and configurable. 

Btw. dprint also supports running other formatters via [exec](https://dprint.dev/plugins/exec/), or pure wasm plugins, so there seems to be a bit of overlap between treefmt and dprint. Maybe there is a possibility for collaboration as it’s also written in rust.

gkze:
Fixes #51


zimbatm (approved):
Looks good, thanks! Nice touch on the wrapper to handle the exposed formatter scenario.

bors merge

PR #84, opened 2023-06-13
Added devShell in module-options build

### LINKED ISSUE #65: [flake-parts] Add a `devShell` option exposing all programs
When using `haskell-flake`, I manually have to pull in the programs into the devShell:

```nix
          devShell = {
            tools = hp: {
              treefmt = config.treefmt.build.wrapper;
            } // config.treefmt.build.programs;
          }
```

cf. https://github.com/srid/haskell-template/blob/fc263b19e4ef02710ffc61fc656aec6c1a873974/flake.nix#L31-L33

---

But we can do this in a cleaner way, using `inputFrom` of mkShell. See https://haskell.flake.page/devshell#composing-devshells

To that end, I'd like a `config.treefmt.build.devShell` option to be added, that I can then pass into `inputsFrom` thereby getting both the wrapper and the underlying programs in the devShell. Happy to open a PR if this is acceptable.

https://github.com/numtide/treefmt-nix/blob/6d8bea2820630576ad8c3a3bde2c95c38bcc471f/module-options.nix#L85-L86

eswar2001:
Resolves #65 .

### DIFF
--- a/module-options.nix
+++ b/module-options.nix
@@ -87,7 +87,10 @@ in
       devShell = mkOption {
         description = "The development shell with treefmt and its underlying programs";
         type = types.package;
-        default = pkgs.mkShell { buildInputs = [ config.build.wrapper ] ++ (lib.attrValues config.build.programs); };
+        readOnly = true;
+        default = pkgs.mkShell {
+          nativeBuildInputs = [ config.build.wrapper ] ++ (lib.attrValues config.build.programs);
+        };
       };
       configFile = mkOption {
         description = ''
--- a/module-options.nix
+++ b/module-options.nix
@@ -84,6 +84,11 @@ in
 
     # Outputs
     build = {
+      devShell = mkOption {
+        description = "The development shell with treefmt and its underlying programs";
+        type = types.package;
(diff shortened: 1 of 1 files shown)


module-options.nix
```diff
@@ -84,6 +84,11 @@ in
 
     # Outputs
     build = {
+      devShell = mkOption {
+        description = "The development shell with treefmt and its underlying programs";
+        type = types.package;
+        default = pkgs.mkShell { buildInputs = [ config.build.wrapper ] ++ (lib.attrValues config.build.programs); };
```
srid, line 90:
```suggestion
        default = pkgs.mkShell { nativeBuildInputs = [ config.build.wrapper ] ++ (lib.attrValues config.build.programs); };
```

(So you get shell completion, as @roberth says)

You probably also want to set `readonly = true;` in this option.

srid (changes requested):

srid (approved):
LGTM

zimbatm:
It's fine, I'll squash the commits. Thanks for the contrib!

PR #86, opened 2023-06-23
Fix option docs rendering

roberth:
devShell default value could not be rendered.
As a readOnly option, it only allows a single definition, so the priority does not matter.
The simplest solution is to make the definition not an option default so that its value is not rendered.

One day I'll add CI for the docs; need to work on the OSS use case a bit.


zimbatm:
thanks, I thought I already had the CI check for that

zimbatm:
bors merge

PR #88, opened 2023-06-28
document flake check with vanilla flake

### LINKED ISSUE #81: Flake check not failing when there is unformatted code
**Describe the bug**

I have unformatted code in the sources, which gets formatted correctly with `nix fmt`, but `nix flake check` does not fail.

**To Reproduce**

1. have unformatted code
2. run `nix flake check`

**Expected behavior**

`nix flake check` should fail

**System information**

I'm using the (currently) most recent version of treefmt-nix (rev 4e92552731aca44ece86731ec4c7848a2c46aa67) on x86_64-linux

**Additional context**

This is my first time using treefmt-nix, so it may very well be the case that I am missing something or my configuration is wrong, I'm not very experienced yet in using nix. But as far as I understood the flake example in the README should also automatically configure the flake check, correct? This is my current minimal configuration:
```
      formatter.x86_64-linux = treefmt-nix.lib.mkWrapper
        nixpkgs.legacyPackages.x86_64-linux
        {
          projectRootFile = "flake.nix";
          programs.nixpkgs-fmt.enable = true;
          programs.stylua.enable = true;
        };

```
It works for `nix fmt`, but not for `nix flake check`.

zimbatm:
Fixes #81


zimbatm:
bors merge

PR #91, opened 2023-07-09
Docs

roberth:
I think the checks are decent now, especially with the `stubPkgs`.
The `flakeModule` isn't covered by the local checks, but the alternative is adding a dependency, which I don't think is a good trade-off because of it leaking to user lock files, https://github.com/NixOS/nix/issues/7730. I think it's good enough.


zimbatm:
thanks!

bors merge

PR #97, opened 2023-07-31
Allow indent_size to be unset for shfmt

### LINKED ISSUE #96: `programs.shfmt.indent_size` breaks shfmt config
Passing any formatting option to `shfmt` causes it to ignore the EditorConfig-based formatting config (see “If any parser or printer flags are given to the tool, no EditorConfig files will be used” in [the README](https://github.com/patrickvane/shfmt#user-content-description).

treefmt-nix will always pass `-i` to `shfmt`, causing it to ignore the actual configuration. There is no option to “unset” `programs.shfmt.indent_size`.

I think `indent_size` should allow (and default to) `null`, only passing `-i` if explicitly requested.

sellout:
So the defaults don’t get in the way of reading the config file.

Fixes #96.

### DIFF
--- a/programs/shfmt.nix
+++ b/programs/shfmt.nix
@@ -7,8 +7,8 @@ in
     enable = lib.mkEnableOption "shfmt";
     package = lib.mkPackageOption pkgs "shfmt" { };
     indent_size = lib.mkOption {
-      type = lib.types.int;
-      default = 2;
+      type = lib.types.nullOr lib.types.int;
+      default = null;
       example = 4;
       description = lib.mdDoc ''
         Sets the number of spaces to be used in indentation. Uses tabs if set to zero.
@@ -19,7 +19,11 @@ in
   config = lib.mkIf cfg.enable {
     settings.formatter.shfmt = {
       command = cfg.package;
-      options = [ "-i" (toString cfg.indent_size) "-s" "-w" ];
+      options =
+        (if isNull cfg.indent_size
+        then [ ]
+        else [ "-i" (toString cfg.indent_size) ])
+        ++ [ "-s" "-w" ];
       includes = [ "*.sh" ];
     };
   };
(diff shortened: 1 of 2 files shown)


programs/shfmt.nix
```diff
@@ -7,8 +7,8 @@ in
     enable = lib.mkEnableOption "shfmt";
     package = lib.mkPackageOption pkgs "shfmt" { };
     indent_size = lib.mkOption {
-      type = lib.types.int;
-      default = 2;
+      type = lib.types.nullOr lib.types.int;
+      default = null;
```
zimbatm, line 11: Changing the default would make this a breaking change for the users.

```suggestion
      default = 2;
```

programs/shfmt.nix
```diff
@@ -19,7 +19,11 @@ in
   config = lib.mkIf cfg.enable {
     settings.formatter.shfmt = {
       command = cfg.package;
-      options = [ "-i" (toString cfg.indent_size) "-s" "-w" ];
+      options =
+        (if isNull cfg.indent_size
+        then [ ]
+        else [ "-i" (toString cfg.indent_size) ])
```
zimbatm, line 25: shorter version:
```suggestion
        (lib.optionals (!isNull cfg.indent_size) [ "-i" (toString cfg.indent_size) ])
```

zimbatm (commented):
thanks. after this round of review, it should be ready to merge.

programs/shfmt.nix
zimbatm: Let's do the non-breaking change as treefmt-nix has no release mechanism at the moment. Users expect the latest main to be relatively stable.

You can use the option description field to describe shfmt's behaviour.

To do the breaking change, I would first add a nix eval trace to warn existing users that the default is going to change, and tell then what to do. Then wait a couple of months and change the default then.

sellout:
Updated!

zimbatm:
thanks!

PR #102, opened 2023-08-06
Specify full path to formatters

gkze:
`nix flake check` results in a warning stating that usage of package names for commands is deprecated, and a full path to the executable must be specified. Add `${lib.getBin cfg.package}/bin/foo` everywhere where only cfg.package is used to specify the command

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1680945546,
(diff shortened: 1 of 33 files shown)


zimbatm:
Thanks for the PR. The issue is going a bit deeper. The best way to fix this would be to add the `meta.mainProgram` field to the packages on the nixpkgs side.

R-VdP:
I added `mainProgram` for most of these in this PR: https://github.com/NixOS/nixpkgs/pull/247755

For some of them I couldn't immediately find the definition, so I skipped those.

gkze:
Thanks, that makes sense; should we close this out then?

zimbatm:
yeah, thanks for the contrib though

PR #103, opened 2023-08-06
Expose Prettier configuration

gkze:
* Expose Prettier configuration options, taken from
  http://json.schemastore.org/prettierrc
* Wrap Prettier invocation in a command that passes generated
  configuration file from passed options

### DIFF
--- a/programs/prettier.nix
+++ b/programs/prettier.nix
@@ -1,25 +1,275 @@
 { lib, pkgs, config, ... }:
 let
+  inherit (lib) filterAttrsRecursive literalMD mkEnableOption mkOption types;
+
   cfg = config.programs.prettier;
+  configFormat = pkgs.formats.json { };
+
+  # Configuration schema for Prettier, which we generate prettierrc.json with.
+  # Definition taken from: http://json.schemastore.org/prettierrc
+  settingsSchema = {
+    arrowParens = mkOption {
+      description = "Include parentheses around a sole arrow function parameter.";
+      type = types.nullOr (types.enum [ "always" "avoid" ]);
+      example = "always";
+      default = null;
+    };
+    bracketSameLine = mkOption {
+      description = ''
+        Put > of opening tags on the last line instead of on a new line.
+      '';
+      type = types.nullOr types.bool;
+      example = false;
+      default = null;
+    };
+    bracketSpacing = mkOption {
+      description = "Print spaces between brackets";
+      type = types.nullOr types.bool;
+      example = true;
+      default = null;
+    };
+    cursorOffset = mkOption {
+      description = ''
+        Print (to stderr) where a cursor at the given position would move to
+        after formatting. This option cannot be used with --range-start and
+        --range-end.
+      '';
+      type = types.nullOr types.int;
+      example = -1;
+      default = null;
+    };
+    editorconfig = mkOption {
+      description = ''
+        Whether parse the .editorconfig file in your project and convert its 
+        properties to the corresponding Prettier configuration. This
+        configuration will be overridden by .prettierrc, etc.
+      '';
+      type = types.nullOr types.bool;
+      example = false;
+      default = null;
+    };
+    embeddedLanguageFormatting = mkOption {
+      description = ''
+        Control how Prettier formats quoted code embedded in the file.
+      '';
+      type = types.nullOr (types.enum [ "auto" "off" ]);
+      example = "auto";
+      default = null;
+    };
+    endOfLine = mkOption {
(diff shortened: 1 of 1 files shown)


gkze:
Stacking on top of #102 to clear `nix flake check` warnings, happy to split

programs/prettier.nix
```diff
@@ -0,0 +1,282 @@
+{ lib, pkgs, config, ... }:
+let
+  inherit (lib) filterAttrsRecursive literalMD mkEnableOption mkOption types;
+
+  cfg = config.programs.prettier;
+  configFormat = pkgs.formats.json { };
+
+  # Configuration scheam for prettier, which we generate prettierrc.json with.
```
zimbatm, line 8: s/scheam/schema/

programs/prettier.nix
```diff
@@ -0,0 +1,282 @@
...
+      description = "Provide a list of patterns to override prettier configuration.";
+      type = types.nullOr (types.listOf types.attrs);
+      example = { files = [ "*.html" "legacy/**/*.js" ]; options.tabwidth = 4; };
+      default = null;
+    };
+  };
+
+  # We filter out null values in dprint configuration since dprint errors when
+  # they're present
+  mkConfigFile = { filename ? ".prettierrc.json", config ? { } }:
```
zimbatm, line 210: let's use `"prettierrc.json"`. That argument can also be collapsed as it's not used or exposed anywhere.

programs/prettier.nix
```diff
@@ -0,0 +1,282 @@
...
+      type = types.package;
+      default = pkgs.nodePackages.prettier;
+      defaultText = lib.literalExpression "pkgs.nodePackages.prettier";
+      description = lib.mdDoc ''
+        prettier derivation to use.
+      '';
+    };
+
+    # Represents the .prettierrc.json config schema
+    config = configSchema;
```
zimbatm, line 227: "config" is an overloaded term in the nix module system; it's best to use "settings" instead.

programs/prettier.nix
```diff
@@ -0,0 +1,282 @@
...
+    excludes = mkOption {
+      description = "Path / file patterns to exclude for Prettier";
+      type = types.listOf types.str;
+      default = [ ];
+    };
+
+    # Wrapped prettier invocation passing a generated .prettierrc.json
+    # configuration
+    wrapper = mkOption {
+      description = "The prettier package, wrapped with the config file.";
```
zimbatm, line 259: this option should be marked as private as it's not meant to be used directly by users

programs/prettier.nix
```diff
@@ -0,0 +1,282 @@
...
+      defaultText = literalMD "wrapped `prettier` command";
+      default =
+        let
+          configFile = mkConfigFile { config = cfg.config; };
+          x = pkgs.writeShellScriptBin "prettier" ''
+            set -euo pipefail
+            exec ${lib.getBin cfg.package}/bin/prettier --config=${configFile} "$@"
+          '';
+        in
+        (x // { meta = config.package.meta // x.meta; });
```
zimbatm, line 270: How would that interact with users that have a prettierrs.json in their repo? Maybe not wrap the program if the user is not adding any config in nix?

gkze: Good point, I added a conditional in the wrapper to check for settings being empty and using prettier directly if so. Let me know if there is a better way.

zimbatm:
looks good. Can you just squash and rebase to remove the other changes from the PR?

gkze:
Lmk if that works

zimbatm:
thanks. I removed the wrapper as it wasn't necessary in the end. Maybe it's worth applying the same changes to drpint.

gkze:
👍🏻 https://github.com/numtide/treefmt-nix/pull/105

programs/prettier.nix
```diff
@@ -1,25 +1,275 @@
...
+      default = null;
+    };
+    plugins = mkOption {
+      description = ''
+        Add a plugin. Multiple plugins can be passed as separate `--plugin`s.
+      '';
+      type = types.nullOr (types.listOf types.str);
+      example = [ "@prettier/plugin-xml" ];
+      default = null;
+    };
```
srid, line 143: https://github.com/numtide/treefmt-nix/issues/112

drupol:
Hello,

Today I just noticed that using `proseWrap` does not have any single effect. However, to get it working properly I added:

```nix
 prettier = {
   options = [
     "--prose-wrap=always"
   ];
   editorconfig = true;
};
```

PR #104, opened 2023-08-07
Clean up dprint integration

gkze:
Same comments from https://github.com/numtide/treefmt-nix/pull/103 apply so doing it here too


zimbatm:
thanks, can you also fix line 119 where it's still using cfg.config instead of cfg.settings?

zimbatm:
thanks!

gkze:
Thanks for the quick turnaround!

PR #109, opened 2023-08-12
add mypy

Mic92:
Unlike normal formatter type checking always needs to take the whole module into account.
That's why cannot just pass single files to mypy. This shouldn't be a problem because mypy never actually format files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,7 @@ This repo contains a top-level `default.nix` that returns the library helper fun
 * mdformat
 * mdsh
 * mix-format
+* mypy
 * nixfmt
 * nixpkgs-fmt
 * ocamlformat
--- a/checks/default.nix
+++ b/checks/default.nix
@@ -25,7 +25,8 @@ let
             } > "$out/${name}.toml"
           ''
         )
-        programConfigs;
+        # mypy example contains store paths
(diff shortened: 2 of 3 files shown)


examples/formatter-mypy.toml
```diff
@@ -0,0 +1,6 @@
+# Example generated by ../examples.sh
+[formatter."mypy-."]
+command = "sh"
+excludes = []
+includes = ["././**/*.py"]
+options = ["-eucx", "cd \".\"\nexport PYTHONPATH=\"/nix/store/jhflvwr40xbb0xr6jx4311icp9cym1fp-python3-3.10.12/lib/python3.10/site-packages\"\n/nix/store/j8s40x2f2fxw83089109xzyjnwxaiqlf-python3.10-mypy-1.3.0/bin/mypy '.'\n"]
```
Mic92, line 6: Don't know how to get rid of this?

Mic92: Filtered it out for now.

Mic92:
Real world example: https://github.com/numtide/harvest-invoice-calculator/commit/88d3f1da65cc042ae8fc6cf4238cb80eac3ddb13

PR #110, opened 2023-08-12
mypy: add options parameter

Mic92:
Since we are in a shell script we cannot pass flags in a different way


Mic92:
Example: https://github.com/Mic92/dotfiles/pull/786/commits/d004aea1a8be53a66c251f7cefaeedcd7d715e24

malteneuss:
A quite important feature for any bigger mypy usage.

PR #125, opened 2023-10-10
Add support for `nickel format`

gkze:
Have to wrap `nickel` executable because it only accepts one file at a time... I do want to make a PR to Nickel & Topiary to accept multiple file / dir args but want to have something working for now until I do that


gkze:
🤔 not sure what's up with the (seemingly?) unrelated CI failures...

zimbatm:
Hey thanks. Make sure to also run the  ./examples.sh and `treefmt` to generate the examples and update the README.md

zimbatm:
there is also an unrelated error dues to terraform switching to BSL

gkze:
(sorry for the delay) I tried running examples.sh but am getting various errors - ktlint not having `meta.mainProgram` and `muon-0.2.0` being marked broken. What should we do?

gkze:
Ty!

zimbatm:
I had to move the wrapper out of the options. The default value was depending on `config`, and that doesn't work (the order of evaluation is imports > options > config).

gkze:
That's right - thank you for the help! 🙇🏻

PR #134, opened 2023-11-30
feat(programs): add golangci-lint

FollieHiyuki:
This PR adds [golangci-lint](https://golangci-lint.run/) as a builtin formatter. It is actually a Go meta-linter, but can be used to fix the incorrect files (if supported by the sub-linters defined by users).


programs/golangci-lint.nix
```diff
@@ -0,0 +1,25 @@
...
+
+  config = lib.mkIf cfg.enable {
+    settings.formatter.golangci-lint = {
+      command = pkgs.writeShellApplication {
+        name = "golangci-lint-run";
+        runtimeInputs = [
+          cfg.package
+          cfg.goPackage
+        ];
+        text = "${cfg.package}/bin/golangci-lint run";
```
zimbatm, line 20: the program should take a list of arguments, otherwise, it will re-format the whole project every time.
```suggestion
        text = "${cfg.package}/bin/golangci-lint run \"$@\"";
```

FollieHiyuki: Thanks! I applied the suggestion.

examples/formatter-golangci-lint.toml
```diff
@@ -0,0 +1,6 @@
+# Example generated by ../examples.sh
+[formatter.golangci-lint]
+command = "golangci-lint-run"
+excludes = []
+includes = ["*.go"]
+options = []
```
zimbatm, line 6: I just noticed that the example is not super useful because it won't work outside of Nix

programs/golangci-lint.nix
```diff
@@ -0,0 +1,25 @@
...
+    settings.formatter.golangci-lint = {
+      command = pkgs.writeShellApplication {
+        name = "golangci-lint-run";
+        runtimeInputs = [
+          cfg.package
+          cfg.goPackage
+        ];
+        text = "${cfg.package}/bin/golangci-lint run \"$@\"";
+      };
+      includes = [ "*.go" ];
```
zimbatm, line 22:
```suggestion
      command = pkgs.writeShellApplication {
        name = "golangci-lint";
        runtimeInputs = [
          cfg.package
          cfg.goPackage
        ];
        text = "${cfg.package}/bin/golangci-lint \"$@\"";
      };
      options = [ "run" ];
      includes = [ "*.go" ];
```
this should fix the example

FollieHiyuki:
Thanks again for the detailed suggestion! The change should be okay now :)

FollieHiyuki:
Sorry @zimbatm, as I did not take the below problem into account before this PR was merged!

I think the change doesn't work as intended. `golangci-lint` requires all the specified files in the argument to be in the same directory to work (or else we need to specify a list of directories instead, recursively or not via `<dir>/...` syntax).

zimbatm:
Ah bummer. That won't work with treefmt then! Do you mind sending a PR to remove it?

FollieHiyuki:
I created https://github.com/numtide/treefmt-nix/pull/135 to remove `golangci-lint` (basically revert this PR). I'll go back to use it in `nix flake check` then :)

PR #141, opened 2023-12-14
just: add as formatter

katexochen:
Adding just formatting.

I'm new to treefmt, but as far as I understand it requires formatters to be able to take files to format as arguments, which just currently doesn't support. I've opened https://github.com/casey/just/issues/1760 as feature request. Opening this PR as draft while looking for a solution...

I've also updated the README section about adding a new formatter, some commands were outdated.


programs/just.nix
```diff
@@ -0,0 +1,24 @@
...
+    package = lib.mkPackageOption pkgs "just" { };
+  };
+
+  config = lib.mkIf cfg.enable {
+    settings.formatter.just = {
+      command = cfg.package;
+      options = [
+        "--fmt"
+        "--unstable" # formtting is still unstable
+        "--justfile" # takes the file as argument, treefmt will append it
```
zimbatm, line 17: Is it able to take multiple files at once? This is require to respect the treefmt formatter spec.

katexochen: No, it is not. :(

katexochen:
I've added a wrapper around just to handle multiple files, similar to what is described in https://github.com/numtide/treefmt/wiki#shell

zimbatm:
cheers. I had to remove the just example because it contains platform-specific paths. Everything is pushed to the main branch now.

PR #146, opened 2024-01-05
add actionlint

katexochen:
I would like to add [actionlint](https://github.com/rhysd/actionlint) (which is more a linter than a formatter, but I saw that shellcheck seems to be okay, so actionlint should fit, too).

However, I'm facing some problems with the include statements. For some reason, files in the `.github` directory are not included/linted. 

I have provided a test commit that activates actionlint and includes two files that should throw errors, one in `.github`, the other in `github`. Running `nix fmt`, this will only print errors for the non-hidden dir.

@zimbatm is treefmt in general not capable of formatting hidden files?


katexochen:
I found https://github.com/numtide/treefmt/issues/153

zimbatm:
this is better handled by treefmt2, since it can now use `git ls-files` as a backend

katexochen:
> this is better handled by treefmt2, since it can now use `git ls-files` as a backend

Nice, looking forward to v2. :)

Mic92:
v2 was released.

zimbatm:
Waiting for nixos-unstable to become available: https://nixpk.gs/pr-tracker.html?pr=323172

Mic92:
This now also happened.

katexochen:
see https://github.com/numtide/treefmt-nix/pull/192 :)

PR #147, opened 2024-01-06
feat(formatters): Add biome.js as a formatter

huwaireb:
Hi, it's been awhile 👋!

I've only tested it by using it practically and running ./examples.sh, the README instructions seem to be out of date. 
It adheres to the formatter spec as far as I see.

### Relevant Links
[Biome](https://biomejs.dev/)


zimbatm:
:wave: happy new year!

Looks good overall. let me know when it's ready!

huwaireb:
> 👋 happy new year!
> 
> Looks good overall. let me know when it's ready!

Cheers 🥳,

I've adjusted the settings to align more with the biome spec, just kept things I thought were relevant.
As well as, instead of using null I opt'd in for using defaults instead and filtered them if they were not in use to avoid setting the config-path.

PR #150, opened 2024-01-18
keep-sorted: init

katexochen:


zimbatm (commented):
As the author of mdsh, I can't accept this change! Joke aside, it looks good :)

I'm just a bit concerned that the list will get out of sync because that new formatter only enforces ordering, not the content of the list. What solution do you propose for that?

katexochen:
> As the author of mdsh, I can't accept this change! Joke aside, it looks good :)

mdsh is great! :)

> I'm just a bit concerned that the list will get out of sync because that new formatter only enforces ordering, not the content of the list. What solution do you propose for that?

I guess I originally created that commit for testing, but then included it into the PR without further thought some days later. Commit is now removed and I also rebased on main to pick up the flake.lock update.

zimbatm:
hehe, cheers :)

PR #155, opened 2024-01-27
add csharpier

thomaslaich:


programs/csharpier.nix
```diff
@@ -0,0 +1,22 @@
...
+  options.programs.csharpier = {
+    enable = lib.mkEnableOption "csharpier";
+    package = lib.mkPackageOption pkgs "csharpier" { };
+    dotnet-sdk = lib.mkPackageOption pkgs "dotnet-sdk" { };
+  };
+
+  config = lib.mkIf cfg.enable {
+    settings.formatter.csharpier = {
+      command = pkgs.writeShellApplication {
+        name = "csharpier-fix";
```
zimbatm, line 13:
```suggestion
        name = "dotnet-csharpier";
```
Needed so the example works for non-nix users

examples/formatter-csharpier.toml
```diff
@@ -0,0 +1,6 @@
+# Example generated by ../examples.sh
+[formatter.csharpier]
+command = "csharpier-fix"
```
zimbatm, line 3:
```suggestion
command = "dotnet-csharpier"
```

zimbatm (commented):
looks good, minus one comment

programs/csharpier.nix
thomaslaich: Quick question about that: when using csharpier without nix, we would usually install it as a local tool in the project like so: https://csharpier.com/docs/About

That means, from the command line we would use `dotnet csharpier` instead of `dotnet-csharpier`. Should I name it accordingly `dotnet csharpier`, or should I go with your suggestion?

Thanks in advance for clarifying!

zimbatm: Let's use `dotnet-csharpier` as the name. I think it would be weird to call the wrapper script `dotnet`. The downloaded tool is still `dotnet-csharpier`, but they seem to use `dotnet` as a dispatcher.

This is an example file, so hopefully the user will figure out if they have to change it.

thomaslaich: fixed

zimbatm:
thanks!

PR #158, opened 2024-02-18
add jsonfmt

thomaslaich:
- add jsonfmt

## Note
I get the following warning when running `nix fmt` with `jsonfmt` enabled (even though formatting works fine):

```
warning: getExe: Package "jsonfmt-0.5.0" does not have the meta.mainProgram attribute. We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold. If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away. Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar"
```

This is because the package definition is missing the `meta.mainProgram` attribute: https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/development/tools/misc/jsonfmt/default.nix#L34

I think this should be fixed upstream. But I'm a nix beginner, so I'm not sure :)


zimbatm:
sounds good!

You can send a PR upstream to set the meta.mainProgram, I'm sure they will appreciate it :)

PR #163, opened 2024-03-03
Fix elm-format

marc136:
The formatting of the elm-format configuration contained additional whitespace due to which formatting did not pick up on the files.


zimbatm:
thanks, can you also run the ./example.sh script to update that folder?

marc136:
Sorry, missed to update the example.

PR #164, opened 2024-03-03
Configurable includes for elm-format

marc136:
This change allows to pass a list of `includes` to the `elm-format` configuration and is an addition after #163.

I want to add this configuration option because in a project I generate a lot of Elm code from another tool. Those generated files are not checked in into source control and don't fully follow the `elm-format` code formatting.

With this change, I can pass include rules that allow me to only format the files that are actually part of the repository and ignore the generated files.

 ```nix
  programs.elm-format.enable = true;
  programs.elm-format.includes = [ "src/Main.elm" ];
```

Thank you for your time and for creating this tool.  
I'm not sure if the desired approach is to immediately create a PR, or first discuss why I would like to add this change.

I copied the code from the biome configuration and adapted it.

### DIFF
--- a/examples/formatter-elm-format.toml
+++ b/examples/formatter-elm-format.toml
@@ -2,5 +2,5 @@
 [formatter.elm-format]
 command = "elm-format"
 excludes = []
-includes = [" * .elm "]
-options = [" - -yes "]
+includes = ["*.elm"]
+options = ["--yes"]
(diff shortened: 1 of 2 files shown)


marc136:
Sorry, I did not run `nix fmt` on the changes and now force-pushed the change to keep the history clean.

zimbatm:
thanks!

PR #169, opened 2024-03-12
Add `packer fmt` formatter

tomeon:
Due to the fact that `packer fmt` accepts only a single file or directory as an argument, this PR includes a not-particularly-pretty wrapper script that calls out to `packer fmt` once per included source file, and a not-particularly-robust mechanism for distinguishing included source files from `packer fmt` options (basically, trying to plop a bare `--` between options and source file arguments).

Also includes some updates to the "Adding new formatters" section of the README; would be happy to spin these changes into a separate PR if that's preferable.

Thanks in advance!


zimbatm (commented):
sounds good. Do you mind also contacting upstream so it eventually gets fixed?

tomeon:
> Do you mind also contacting upstream so it eventually gets fixed?

Done: hashicorp/packer#12883.

dawidd6:
Isn't this the same as `hclfmt` though?

zimbatm:
probably. I would use `hclfmt` as it follows the treefmt spec. But I haven't used Packer for a while, maybe there is some extra semantic that `packer fmt` understands?

tomeon:
> [M]aybe there is some extra semantic that packer fmt understands?

Allegedly several of the `<hashicorp-tool> fmt` subcommands, including `packer fmt`, have tool-specific behavior, according to Hashicorp employees/contributors [here](https://github.com/hashicorp/vscode-hcl/issues/98#issuecomment-1137569350) and [here](https://github.com/hashicorp/vscode-hcl/issues/98#issuecomment-1875159007).

PR #172, opened 2024-03-27
php-cs-fixer: add configFile option

RTUnreal:
This fixes an issue with php-cs-fixer, where it throws an error about requiring the `--config` argument if multiple files to format are specified.


zimbatm:
thanks!

PR #174, opened 2024-03-30
nimpretty: Add nimpretty

haruki7049:
The nim-lang's repo: https://github.com/nim-lang/Nim
The nim-lang's homepage: https://nim-lang.org


zimbatm:
thanks!

PR #175, opened 2024-04-24
ruff: allow both checking and formatting

pvsr:
Right now when you enable ruff it will either do just linting or just formatting, depending on the value of `programs.ruff.format`. This looks more like a bug rather than a deliberate decision, since I don't know of a reason they should be mutually exclusive. This PR creates separate formatter objects for ruff as a linter and ruff as a formatter, allowing them to be used together. It also switches from the deprecated `ruff {files}` linter invocation to `ruff check {files}`, supported since https://github.com/astral-sh/ruff/releases/tag/v0.0.237 in Jan 2023, and renames `ruff.enable` to `ruff.check` to clarify the distinction between linting and formatting.


zimbatm:
thanks!

PR #176, opened 2024-05-02
choose local that is compatible with macOS

Mic92:


zimbatm (commented):
Isn't this going to change the ordering of some output, making it less reproducible?

Mic92:
Well. There is no such thing as a C.UTF-8 in macOS, so just now it's even less reproducible because of it. Also it throws warnings.

PR #182, opened 2024-06-01
hclfmt: only format .hcl files

zimbatm:
Avoid clashes with the `terraform` formatter who does a better job at formatting the ".tf" files, since it has more domain knowledge there.

### DIFF
--- a/examples/formatter-hclfmt.toml
+++ b/examples/formatter-hclfmt.toml
@@ -2,5 +2,5 @@
 [formatter.hclfmt]
 command = "hclfmt"
 excludes = []
-includes = ["*.hcl", "*.tf"]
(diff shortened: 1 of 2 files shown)


zowoq:
Using hclfmt for `tf` files was useful as terraform has an unfree license.

zimbatm:
In that case you can set `programs.terraform.package = pkgs.opentofu` no?

zowoq:
I dropped the hack we were using and set allowUnfreePredicate instead. https://github.com/nix-community/infra/pull/1296

Mic92:
@zimbatm should we default to opentofu? Presumably it doesn't make a difference for formatting.

zimbatm:
yeah good idea

PR #186, opened 2024-06-12
feat(flake-parts): add `pkgs` module option

shivaraj-bh:


srid (approved):

srid:
I believe @shivaraj-bh added this because treefmt-nix requires a newer nixpkgs, so for projects that still use older nixpkgs, we could solve that by explicitly passing a `pkgs-latest` to treefmt-nix.

cc @roberth (this may be a general thing with all flake-parts modules?)

roberth:
That's very similar to what I've done in pre-commit, https://flake.parts/options/pre-commit-hooks-nix#opt-perSystem.pre-commit.pkgs

In that one, the pre-commit-hooks modules are loaded into a sub-option `settings`, which you don't have here, which I think is good, and that's the only real difference.

We could call it a pattern. :+1:

flake-module.nix
```diff
@@ -21,7 +21,15 @@ in
...
-            modules = (import ./.).all-modules pkgs ++ [{
+            modules = (import ./.).all-modules config.treefmt.pkgs ++ [{
+              options.pkgs = mkOption {
+                type = types.uniq (types.lazyAttrsOf (types.raw or types.unspecified));
+                description = ''
+                  Nixpkgs to use in `treefmt`.
+                '';
+                default = pkgs;
+                defaultText = "`pkgs` (module argument)";
+              };
```
roberth, line 32: Although it _can_ all be done here, I believe it is better to make `default.nix` more aware of the use case of being loaded into a submodule. It doesn't have to know about flake-parts; what I'm suggesting is just to make this use case more visible by adding a `submodule-modules` module.

```suggestion
            modules = (import ./.).submodule-modules ++ [{
              options.pkgs = mkOption {
                defaultText = "`pkgs` (`perSystem` module argument)";
              };
```

`default.nix`
```nix
all-modules = ...;

# treefmt-nix can be loaded into a submodule. In this case we get our `pkgs` from
# our own standard option `pkgs`; not externally.
submodule-modules = 
   [
    {
      options.pkgs = mkOption {
        type = types.uniq (types.lazyAttrsOf (types.raw or types.unspecified));
        description = ''
          Nixpkgs to use in `treefmt`.
         '';
        default = pkgs;
      };
      config._module.args = {
        pkgs = nixpkgs;
      };
    }
    module-options
  ]
  ++ programs.modules
```

`_module.args.lib` doesn't do anything, because `lib` already behaves like a `specialArgs` argument, so that it can be used when computing `imports`.

roberth: It's really just a refactor though; it behaves exactly the same, so let's do whichever solution the maintainers prefer.

roberth (approved):

flake-module.nix
zimbatm: That sounds good. @shivaraj-bh do you mind integrating Robert's changes?

shivaraj-bh: Will do!

shivaraj-bh: Done here: https://github.com/numtide/treefmt-nix/pull/186/commits/3123e3763be68eee59dbc106a8b8c31bb73e06ec

default.nix
```diff
@@ -18,6 +18,32 @@ let
...
+          types;
+      in
+      {
+        options.pkgs = mkOption {
+          type = types.uniq (types.lazyAttrsOf (types.raw or types.unspecified));
+          description = ''
+            Nixpkgs to use in `treefmt`.
+          '';
+          default = pkgs;
+          defaultText = "`pkgs` (module argument)";
```
roberth, line 37: Could you move these two back to the flake-parts module?
That will fix the accidental circular definition of `pkgs`, and you can also make the default text more descriptive:
```nix
defaultText = "`pkgs` (module argument of `perSystem`)";
```

Option _declarations_ are also merged, as long as the option attribute don't conflict.

PR #191, opened 2024-07-01
terraform: default on opensource opentofu for formatting

Mic92:
Than we don't have to compile terraform from source.


zimbatm:
good idea. we can also drop the allowUnfree in that case.

Mic92:
CI disagrees we still need unfree.

PR #192, opened 2024-07-01
Update to treefmt v2

katexochen:


programs/gdformat.nix
```diff
@@ -5,7 +5,7 @@ in
 {
   options.programs.gdformat = {
     enable = lib.mkEnableOption "gdformat";
-    package = lib.mkPackageOption pkgs "gdtoolkit" { };
+    package = lib.mkPackageOption pkgs "gdtoolkit_4" { };
```
katexochen, line 8: > gdtoolkit has been renamed to gdtoolkit_3 to distinguish from version 4

Unsure if we should support both gdtoolkit_3/4, maybe @haruki7049 can give some input on that.

programs/nixfmt-classic.nix
```diff
@@ -0,0 +1,17 @@
+{ lib, pkgs, config, ... }:
+let
+  cfg = config.programs.nixfmt-classic;
+in
+{
+  options.programs.nixfmt-classic = {
+    enable = lib.mkEnableOption "nixfmt-classic";
```
katexochen, line 7: > nixfmt was renamed to nixfmt-classic. The nixfmt attribute may be used for the new RFC 166-style formatter in the future, which is currently available as nixfmt-rfc-style

Renaming nixfmt -> nixfmt-classic is a breaking change.
cc @SebTM who introduced nixfmt linter.

zimbatm: Given that both nixfmt-classic and nixfmt-rfc-style are CLI compatible, and that nixfmt-classic will probably disappear, I propose:
1. Keep `nixfmt` as the module name
2. Use `nixfmt-rfc-style` as the default package.
3. Deprecate the `nixfmt-rfc-style` module.
And if people want to use `nixfmt-classic`, they can set it as the package.

programs/gdformat.nix
zimbatm: As long as it can be overwritten by the user, we don't need to support all of the options.

programs/nixfmt-classic.nix
katexochen: Added a warning to `nixfmt-rfc-style`, set nixfmt default package to `nixfmt-rfc-style`.

programs/gdformat.nix
haruki7049: I, too, don't think it is necessary to support all options if the user can override the options...

zimbatm:
thanks!

PR #193, opened 2024-07-04
yamlfmt: add settings option

katexochen:


programs/yamlfmt.nix
```diff
@@ -6,11 +6,96 @@ in
...
+              indentless_arrays
+              drop_merge_tags
+              pad_line_comments
+              trim_trailing_whitespace
+              eof_newline;
+          };
+          stringify = v: if lib.isBool v then lib.boolToString v else toString v;
+        in
+        lib.mapAttrsToList
+          (n: v: "-formatter=${n}=${stringify v}")
```
zimbatm, line 97: I think it would be best to generate the config file using the "settings" pattern and then pass it here with the `-conf` argument (if any settings are being set). The main reason is that the kernel limits the amount of argument a process can take, and we are using most of them for passing the list of files to format.

katexochen: Makes sense, will rework as soon as I find the time. :)

zimbatm: Are you still interested in working on this?

katexochen: Sorry, didn't find the time. I'll get a coffee and sit down somewhere to finish this now. :)

zimbatm (approved):
Thanks!

PR #194, opened 2024-07-07
exclude some file types by default

Mic92:
Since we now have treefmt warn about unformatted files it might make sense to exclude some files we commonly never want to reformat


Mic92:
@mergify rebase

Mic92:
Unsure why the github status is missing on some checks, but CI is green.

PR #195, opened 2024-07-07
{shellcheck,shfmt}: also format .bash and .envrc files

Mic92:


zimbatm:
sounds good, the examples just need to be re-generated

Mic92:
Done.

Kamillaova:
@Mic92 
Is there any real reason why you included .envrc for shellcheck? and also why you added `*.envrc` instead of `.envrc`?

yes, direnv files are shell scripts, but
1. usually `.envrc` doesn't has shebang (and shellcheck wants it)
2. I think there are direnv modules that can add environment variables, and maybe it might be useful to use them in `.envrc`, but shellcheck can't traverse these modules, and because of that it will complain about missing environment variables (and will require warning exclusion/`?:` assertion)

Mic92:
You don't need a shebang for shellcheck to work: https://github.com/Mic92/dotfiles/blob/0e6e9568947c0541b596e00eee783fca02c80cc4/.envrc#L1

PR #196, opened 2024-07-10
mypy: add package option

phaer:


programs/mypy.nix
```diff
@@ -4,6 +4,7 @@
     enable = lib.mkEnableOption "mypy";
     directories = lib.mkOption {
       description = "Directories to run mypy in";
+      package = lib.mkPackageOption pkgs "mypy" { };
```
Mic92, line 7: You need to move this option a few lines up.

phaer: Haha, of course. Thanks for the catch, I was too confident in just putting this in quickly without verification and of course choose the wrong line.

PR #199, opened 2024-07-14
feat(programs): add fish_indent

terlar:
This adds support for formatting fish files via fish_indent bundled with fish. Not sure about the naming, if it should be how the tool is named or if it should be just "fish".


terlar:
Unfortunately `fish_indent` write files every time currently so had to create a wrapper.

`fish_indent --check` will return the files that needs to change to stderr.

I created an issue for fish-shell:
https://github.com/fish-shell/fish-shell/issues/10616

zimbatm:
Thanks. I'll merge this, but can you send up a follow-up PR to remove the wrapper once fish-shell has released a new version?

PR #203, opened 2024-07-19
actionlint: revert including action files

katexochen:
Sorry, I think I misunderstood the way actionlint consumes the actions.yml files. It looks like passing the files directly isn't intended for actions.

> actionlint still focuses on checking workflow files. So there is no way to directly specify action.yml as an argument of actionlint command.

https://github.com/rhysd/actionlint/releases/tag/v1.7.0


zimbatm:
ok, no worries

PR #208, opened 2024-08-03
mypy: allow adding custom paths to the shell

DavHau:
Problem this fixes:
Adding in-tree python dependencies via `extraPythonPackages` leads to constant re-builds, as the treefmt wrapper becomes dependant on ones own code.

This adds a new option `extraPythonPaths` which allows adding local projects to PYTHONPATH relative to $PROJECT_ROOT


DavHau:
I noticed that this doesn't quite work yet.

DavHau:
> I noticed that this doesn't quite work yet.

Fixed

module-options.nix
```diff
@@ -227,10 +227,10 @@ in
               # `treefmt --fail-on-change` is broken for purs-tidy; So we must rely
               # on git to detect changes. An unintended advantage of this approach
               # is that when the check fails, it will print a helpful diff at the end.
-              PRJ=$TMP/project
-              cp -r ${self} $PRJ
-              chmod -R a+w $PRJ
-              cd $PRJ
+              export PROJECT_ROOT=$TMP/project
```
zimbatm, line 230: Let's name it PRJ_ROOT to follow the [prj-spec](https://github.com/numtide/prj-spec)

programs/mypy.nix
```diff
@@ -23,6 +23,17 @@
                 example = lib.literalMD ''[ pkgs.python3.pkgs.requests ]'';
                 description = "Extra packages to add to PYTHONPATH";
               };
+              extraPythonPaths = lib.mkOption {
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
+                example = [ "$PROJECT_ROOT/path/to/my/module" ];
```
Mic92, line 29:
```suggestion
                example = [ "$PRJ_ROOT/path/to/my/module" ];
```

programs/mypy.nix
```diff
@@ -23,6 +23,17 @@
...
+                description = ''
+                  Extra paths to add to PYTHONPATH relative.
+
+                  Add this to your shellHook first:
+                    export PROJECT_ROOT=$(git rev-parse --show-toplevel)
```
Mic92, line 34:
```suggestion
                    export PRJ_ROOT=$(git rev-parse --show-toplevel)
```

module-options.nix
Mic92, line 230:
```suggestion
              export PRJ_ROOT=$TMP/project
```

module-options.nix
```diff
@@ -227,10 +227,10 @@ in
...
+              cp -r ${self} $PROJECT_ROOT
+              chmod -R a+w $PROJECT_ROOT
+              cd $PROJECT_ROOT
```
Mic92, line 233:
```suggestion
              cp -r ${self} $PRJ_ROOT
              chmod -R a+w $PRJ_ROOT
              cd $PRJ_ROOT
```

module-options.nix
Mic92: Actually relative paths work in this project as well. I don't think we need this environment variable.

Mic92:
@mergify queue

Mic92:
@mergify queue

PR #213, opened 2024-08-10
chore(ruff): split into ruff-format and ruff-check

zimbatm:
This fixes ./example.sh and `nix flake check`. They depends on having .enable options.


programs/ruff-format.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  imports = [
+    (lib.mkRenamedOptionModule [ "programs" "ruff" "format" ] [ "programs" "ruff-format" "enable" ])
+  ];
+
+  options.programs.ruff-format = {
+    enable = lib.mkEnableOption "ruff formatter" // {
+      description = ''
+        Whether to enable the Ruff formatter, an extremely fast Python code formatter
+        designed as a drop-in replacement for Black.
```
declension, line 16: I'd question this bit somewhat, but I see what you mean

declension:
Does this also fix all the 
```
trace: Obsolete option `programs.ruff.enable' is used. It was renamed to `programs.ruff.check'.
```
we now see everywhere?

zimbatm:
the constant warning was fixed separated (#214 and #218)

Issue #219, opened 2024-08-10
flake-parts module is satisfied neither with nor without `projectRootFile`


emilazy:
The flake-parts module sets `projectRoot`, so if you set `projectRootFile` you get `treefmt: error: --tree-root and --tree-root-file can't be used together`. However, if you don’t set it, you get ``error: The option `perSystem.x86_64-darwin.treefmt.projectRootFile' is used but not defined.``.

emilazy:
I’m not actually sure where the CLI tool is getting `--tree-root` to complain about it. The generated configuration file doesn’t set anything like that, and it’s not passed as a parameter directly.

emilazy:
Ah, I think this might actually be an unfortunate devshell interaction, because `PRJ_ROOT` is set…

In any case, presumably `projectRootFile` should be optional.

PR #220, opened 2024-08-11
fix: unset PRJ_ROOT to prevent impure behavior

GaetanLepage:
When the `PRJ_ROOT` environment variable is set, [`treefmt` will use it to populate the `--tree-root` option](https://github.com/numtide/treefmt/blob/31e363dbc7d99e83e9924be21e9e169bbfea98f3/docs/usage.md?plain=1#L24).
Also, both `--tree-root` and `--tree-root-file` options must not be set at the same time, otherwise leading to:
```
treefmt: error: --tree-root and --tree-root-file can't be used together
```

As the tree-fmt module is explicitly setting `--tree-root-file`, I don't see a justification to let `PRJ_ROOT` leak in this script.

### DIFF
--- a/module-options.nix
+++ b/module-options.nix
@@ -180,6 +180,7 @@ in
               else
                 ''
                   set -euo pipefail
+                  unset PRJ_ROOT
                   exec ${config.package}/bin/treefmt \
                     --config-file=${config.build.configFile} \
                     --tree-root-file=${config.projectRootFile} \


GaetanLepage:
To give a bit more context, I faced this issue while using [devshell](https://github.com/numtide/devshell) which automatically sets the `$PRJ_ROOT` environment variable.

The combination of `devshell` and `treefmt-nix` thus leads to the aforementioned error message.

emilazy:
cc https://github.com/numtide/treefmt-nix/issues/219

I think the flake-parts module may still be broken after this, cf. https://github.com/numtide/treefmt-nix/blob/c9f97032be6816fa234f24803b8ae79dc7753a91/flake-module.nix#L43-L50.

emilazy:
Or maybe not? `projectRoot` is only used for the check.

Anyway I was thinking that making `projectRootFile` optional might be a good idea too.

GaetanLepage:
Well, I don't know where this should be fix.
But the formatting script, as it is generated right now, should definitely ignore the `PRJ_ROOT` variable. At best, the latter is not set and thus ignored, but at worst, it generates the error.

Of course, there might be a cleaner solution.

emilazy:
Right. The potential alternative, I guess, is to make `projectRootFile` optional and omit `--tree-root-file` when `PRJ_ROOT` is set. Not sure what would be best.

(This needs rebasing, by the way, I think.)

MattSturgeon:
> The potential alternative, I guess, is to make `projectRootFile` optional and omit `--tree-root-file` when `PRJ_ROOT` is set. Not sure what would be best.

I don't see why we would want impure environment variables leaking through to treefmt? Isn't it possible that `PRJ_ROOT` could end up being set to some arbitrary value that is unrelated to what was configured in the treefmt-nix flake module? The environment shouldn't be able to override something explicitly configured in a module and defined in the wrapper script.

emilazy:
Well, I think searching for the project root is kind of inherently impure to begin with. But sure, that’s fair enough. (`projectRoot` being an option that’s only used for the flake check, rather than setting `--tree-root` similarly to `projectRootFile`, is pretty confusing in itself, but that’s another matter.)

zimbatm:
This is a bug in treefmt. Env vars should not take precedence over CLI args.

zimbatm:
https://github.com/alecthomas/kong/issues/446

zimbatm:
I'll merge this for now

PR #224, opened 2024-08-21
add toml-sort

nim65s:
Hi,

This add https://github.com/pappasam/toml-sort, which was added in nixpkgs in https://github.com/NixOS/nixpkgs/pull/319793


zimbatm:
thanks!

PR #235, opened 2024-09-18
Add meson fmt

RossSmyth:
There is supported for muon, but not meson's formatter. So here it is. I tested this config out on my work repo and it worked exactly as expected.


zimbatm:
thanks!

PR #238, opened 2024-09-20
feat: add nixfmt-classic

anntnzrb:
Adds the now old `nixfmt-classic` formatter, even though it's currently unmaintained and has been superseded by the `nixfmt-rfc-style` standard.

I was initially hesitant about this but some users may still prefer it, the option for using it is offered, with a clear warning. I was also a bit unsure about the wording in the warning message, so any feedback on that would be appreciated.


zimbatm:
Isn't it enough to set `programs.nixfmt.package = pkgs.nixfmt-classic`?

anntnzrb:
> Isn't it enough to set `programs.nixfmt.package = pkgs.nixfmt-classic`?

In short, yes. However, since `nixfmt-rfc-style` option is getting removed in the future, perhaps there could be an option for this at `nixfmt`.

In reality, this would only serve to avoid creating a `settings. formatter` block.

PR #240, opened 2024-09-22
rustfmt: add skip children

ProfDoof:
Currently, rustfmt will recursively format module out of line files which is counter to the standards stated for formatters. I am hoping to get a response from the rustfmt folks in particular as to the expected way to handle this but this is the solution I found that works. 

If they get back to me, I will update this PR. Otherwise, this should prevent rustfmt from going off and doing it's own thing. Probably not a huge deal but did seem relevant.

### DIFF
--- a/examples/formatter-rustfmt.toml
+++ b/examples/formatter-rustfmt.toml
@@ -3,4 +3,4 @@
 command = "rustfmt"
 excludes = []
 includes = ["*.rs"]
-options = ["--edition", "2021"]
+options = ["--edition", "2021", "--config", "skip_children=true"]
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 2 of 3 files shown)


ProfDoof:
I'll go ahead and run ./examples.sh shortly. Just takes a while on my machine.

zimbatm:
thanks. Can you share the link to the upstream issue so I can better understand the problem?

ProfDoof:
I'm not sure I understand the request. There's not an issue in rustfmt. Do you mean the discussion where I ask about expected behavior?

https://github.com/rust-lang/rustfmt/discussions/6345

ProfDoof:
Ok, I can't run examples.sh on my machine right now as deno is not building for x86_64-darwin which I'm on. https://hydra.nixos.org/build/271827764/nixlog/4

I am going to try to figure that out but until then, if examples.sh needs to be run, it'll probably need to be run by someone else.

PR #242, opened 2024-09-24
meson: format all meson files

### LINKED ISSUE #241: Meson formatters do not recurse into child directories
**Describe the bug**

**To Reproduce**

1. Create a top-level `meson.build` file
2. Create a `meson.build` file with any child directory i.e. `mkdir foo && touch foo/meson.build`
3. Run formatter

For example mine:
image

It does format the top-level one though.

**Expected behavior**

Formats all files. I tried fixing this by using the rules here:
https://github.com/numtide/treefmt-nix/compare/main...RossSmyth:treefmt-nix:main

As that globs the files in my shell. But it still outputs that the files are not formatted.

**System information**

35dfece10c642eb52928a48bee7ac06a59f93e9a
https://github.com/NixOS/nixpkgs/commit/a1d92660c6b3b7c26fb883500a80ea9d33321be2

**Additional context**
I'll fix it if desired, I just need to know the globbing rules

RossSmyth:
Fixes #241 

Previously the formatters only globbed the top-level files. I tried several different combos and found these work. The other globs I tried:

`*/meson.build`
`**/meson.build`

Did not work. But these seem to.

### DIFF
--- a/examples/formatter-meson.toml
+++ b/examples/formatter-meson.toml
@@ -2,5 +2,5 @@
 [formatter.meson]
 command = "meson"
 excludes = []
-includes = ["meson.build", "meson.options", "meson_options.txt"]
+includes = ["meson.build", "meson.options", "meson_options.txt", "*/meson.build", "*/meson.options", "*/meson_options.txt"]
 options = ["fmt", "-i"]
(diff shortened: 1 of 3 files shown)


zimbatm:
thanks!

PR #252, opened 2024-10-24
stylua: add options: `settings`, `includes`, `excludes`; add to `meta.maintainers`

sebaszv:
Added `settings` options, which produce an optional `stylua.toml` config
file for `stylua` to use.
This was inspired by the `programs/prettier.nix` module which has a
similar implementation for its `prettierrc.json` for `prettier` through
its `settings` options.

`includes` and `excludes` options were added to easily override the
default if desired, rather than having to go through
`settings.formatter.stylua` separately.

I also added myself to `meta.maintainers`.

### DIFF
--- a/programs/stylua.nix
+++ b/programs/stylua.nix
@@ -165,7 +165,7 @@ let
       null;
 in
 {
-  meta.maintainers = [ ];
+  meta.maintainers = [ "sebaszv" ];
 
   options.programs.stylua = {
     enable = lib.mkEnableOption "stylua";
--- a/programs/stylua.nix
+++ b/programs/stylua.nix
@@ -1,6 +1,168 @@
 { lib, pkgs, config, ... }:
 let
+  inherit (lib.types)
+    bool
+    int
+    str
+    enum
+    listOf
+    nullOr;
+
   cfg = config.programs.stylua;
+  configFormat = pkgs.formats.toml { };
+
+  /*
+   * The schema and descriptions were taken from the StyLua README
+   * on the project's GitHub page:
+   * <https://github.com/JohnnyMorganz/StyLua/blob/main/README.md>
+   */
+  settingsSchema = {
+    column_width = lib.mkOption {
+      description = ''
+        Approximate line length for printing.
+
+        Used as a guide for line wrapping -
+        this is not a hard requirement:
+        lines may fall under or over the limit.
+      '';
+      type = nullOr int;
+      example = 120;
+      default = null;
+    };
+
+    line_endings = lib.mkOption {
+      description = ''
+        Line endings type.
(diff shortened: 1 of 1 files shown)


programs/stylua.nix
```diff
@@ -1,19 +1,200 @@
...
+      filterOutEmptyAttrs =
+        lib.filterAttrsRecursive
+          (_: v: v != { });
+
+      settings = filterOutEmptyAttrs (filterOutNull cfg.settings);
+    in
+    if settings != { } then
+      configFormat.generate "stylua.toml" settings
+    else
+      null;
```
zimbatm, line 165:
```suggestion
    configFormat.generate "stylua.toml" settings;
```

programs/stylua.nix
```diff
@@ -1,19 +1,200 @@
...
+      type = listOf str;
+      default = [ ];
+    };
   };
 
   config = lib.mkIf cfg.enable {
     settings.formatter.stylua = {
       command = cfg.package;
-      includes = [ "*.lua" ];
+      options = lib.mkIf (settingsFile != null) [
```
zimbatm, line 191:
```suggestion
      options = lib.mkIf (settings != { }) [
```

programs/stylua.nix
```diff
@@ -185,7 +185,7 @@ in
   config = lib.mkIf cfg.enable {
     settings.formatter.stylua = {
       command = cfg.package;
-      options = lib.mkIf (settingsFile != null) [
+      options = lib.mkIf (settings != { }) [
```
sebaszv, line 188: I tested this updated version and it breaks the module.

Running `./examples.sh` confirms this. I'm not sure if I am missing something here.

programs/stylua.nix
```diff
@@ -159,10 +159,7 @@ let
 
       settings = filterOutEmptyAttrs (filterOutNull cfg.settings);
     in
-    if settings != { } then
-      configFormat.generate "stylua.toml" settings
-    else
-      null;
+    configFormat.generate "stylua.toml" settings;
```
sebaszv, line 162: Is it standard to always generate a file regardless?
The intent was to avoid generating a config file at all unless actually needed.
I'm not sure if I've missed something.

sebaszv:
I reverted the changes because they were breaking and some of my projects are using the branch.

programs/stylua.nix
zimbatm: Yeah, Nix is lazily evaluated, so it would only generate the file if the branch is also evaluated.

programs/stylua.nix
zimbatm: I took another look, and your code is good enough.

PR #256, opened 2024-11-06
google-java-format: add `aospStyle`, `includes`, `excludes`

sebaszv:
These are simple additions.

- `aospStyle`: For those that need 4-spaced indentation, rather than 2-spaced.
- `includes` && `excludes`: Exposed through `programs.google-java-format` rather
  than only through `settings.formatter.google-java-format`.

I also added myself to the maintainers list.


sebaszv:
@zimbatm Not sure if this was missed.

zimbatm (approved):
yep, thanks!

PR #258, opened 2024-11-12
feat: sqlfluff support

o-az:


o-az:
current ci failure is because dialect option is not being passed to sqlfluff but it's required

zimbatm:
Our test framework doesn't support options that have no defaults, so I put "ansi" as the default dialect

o-az:
perfect thank you.

o-az:
anything else needed to get this through? Thank you

zimbatm (approved):
LGTM, thanks

PR #260, opened 2024-11-15
feat: update nixpkgs input

brianmcgee:
Brings in treefmt v2.1.0

Signed-off-by: Brian McGee <brian@bmcgee.ie>


zimbatm:
it looks like Zig if broken on macOS

PR #268, opened 2024-11-26
treefmt config: make global settings global

kingarrrt:
Put global settings at the top level of the generated toml.

Depends on #267. Needs more work.


zimbatm:
I'm not sure what problem this solves?

PR #269, opened 2024-12-01
escape names in formatter names

Mic92:
We used to allow . and / in formatter names but no longer.


zimbatm:
This might be a regression in treefmt itself. What do you think @brianmcgee ?

brianmcgee:
I've created https://github.com/numtide/treefmt/issues/492 to follow up

PR #275, opened 2024-12-09
🆙 Include all opentofu-compatible files for Terraform

### LINKED ISSUE #274: Terraform formatter should include all formattable files
**Describe the bug**

`nix fmt` doesn't format `.tfvars` or `.tftest.hcl` when it could

From opentofu's own error message:

> Only .tf, .tfvars, and .tftest.hcl files can be processed with tofu fmt

**To Reproduce**

1. Have a reasonble sized TF repo, including some `.tfvars` files and maybe `.tftest.hcl`
2. Make sure one of those tfvars has crazy indentation, or something.
3. Configure `treefmt.programs.terraform.enable = true;`
4. Run `nix fmt`

**Expected behavior**
It formats everything it can by default, including the tfvars etc

**System information**
`treefmt2` via nixpkgs 24.11

declension:
Fixes #274

### DIFF
--- a/examples/formatter-terraform.toml
+++ b/examples/formatter-terraform.toml
@@ -2,5 +2,5 @@
 [formatter.terraform]
 command = "tofu"
 excludes = []
-includes = ["*.tf"]
+includes = ["*.tf", "*.tfvars", "*.tftest.hcl"]
 options = ["fmt"]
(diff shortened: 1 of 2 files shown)


declension:
> thanks, can you run `./examples.sh` and push the result?

Thanks - I was just doing that actually! Should be updated now

PR #281, opened 2024-12-20
perltidy: init

haruki7049:
# Changes
- Init `/programs/perltidy.nix`
- Execute `/example.sh`, then it generates `/examples/formatter-perltidy.toml`


haruki7049:
DONE

programs/perltidy.nix
```diff
@@ -0,0 +1,56 @@
...
+    settings.formatter.perltidy = {
+      command = cfg.package;
+
+      options =
+        if cfg.perltidyrc == null then
+          cfg.options
+        else
+          [
+            "-pro=${cfg.perltidyrc}"
+          ] ++ cfg.options;
```
zimbatm, line 49:
```suggestion
        lib.optional (cfg.perltidyrc != null) "-pro=${cfg.perltidyrc}"
        ++ cfg.options;
```
Something like that shows clearer intent

zimbatm (commented):
Small nit, the rest looks good to me

zimbatm:
thanks for the contrib!

PR #285, opened 2024-12-30
feat: add 'goimports' formatter

gabyx:
Thanks for merging.

P.S.: Great f**** project. Global formatting is the way to goooo!


jaredmontoya:
@gabyx is it possible to also add goimports-reviser? I tried to do so but it seems buggy:
https://github.com/jaredmontoya/treefmt-nix/tree/add-goimports-reviser

gabyx:
@jaredmontoya Hi, :). Not sure, what do you experience that it makes it buggy? Maybe the formattee does not behave to the rules?

PR #288, opened 2024-12-31
standardize program options with mkFormatterModule

zimbatm:
A common complaint / observation is that treefmt-nix config tend to be a mix between enabled programs, and treefmt settings to override includes/excludes/priority. Some programs have those options available directly, but not all of them.

In order to fix that, this commit introduces `mkFormatterModule`, a function that generates a regular module structure based on a few parameters. And that support the includes / excludes / priority out of the box.

A few of the programs are already converted to that format. Ideally all of them are getting converted soon.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -7,6 +7,71 @@ let
   # Program to formatter mapping
   programs = import ./programs.nix;
 
+  mkFormatterModule =
+    {
+      name,
+      package ? name,
+      mainProgram ? null,
+      args ? [ ],
+      includes ? [ ],
+      excludes ? [ ],
+    }:
+    {
+      pkgs,
+      lib,
+      config,
+      ...
+    }:
+    let
+      cfg = config.programs.${name};
+    in
+    {
+      options.programs.${name} = {
+        enable = lib.mkEnableOption name;
+
+        package = lib.mkPackageOption pkgs package { };
+
+        includes = lib.mkOption {
+          description = "Path / file patterns to include";
+          type = lib.types.listOf lib.types.str;
+          default = includes;
+        };
+
+        excludes = lib.mkOption {
+          description = "Path / file patterns to exclude";
+          type = lib.types.listOf lib.types.str;
+          default = excludes;
+        };
+
+        priority = lib.mkOption {
+          description = "Priority";
+          type = lib.types.nullOr lib.types.int;
+          default = null;
+        };
+      };
+
+      config = lib.mkIf cfg.enable {
+        settings.formatter.${name} =
+          {
+            command = lib.mkDefault (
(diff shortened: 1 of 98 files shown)


brianmcgee (approved):

zimbatm:
Ok this is ready. I was able to convert all the formatters except mypy that does some funky stuff.

Looking for feedback from all the maintainers since this changes all the modules. /cc @anntnzrb @sebaszv @haruki7049 @gabyx @SigmaSquadron @adam-gaia @RossSmyth @JakobLichterfeld

phaer (approved):
Very nice! Seems to work, much less boilerplate code but same features and a cleaner interface.

Just curious if you did use any tooling for the module changes or did that all "by hand"?

sebaszv:
This was something I did pick up on as being an annoyance. Glad this is standardised now. LGTM 👍

anntnzrb:
✔️ Great refactor, contributions should be even easier now.

phaer:
Bug reports are more useful if they contain info on what breaks and how that breakage might be reproducible. :)

It seems to be broken indeed, runs into infinite recursion while evaluating the module arg

```
       … while evaluating the module argument `mkFormatterModule' in "/nix/store/qa
r0bxqnnbl6j6i3yh34h73xafah1bi5-source/programs/actionlint.nix":

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: infinite recursion encountered
       at /nix/store/j0jlb33bm31w347w43051hbffgh616jr-source/lib/modules.nix:506:28
:
          505|         builtins.addErrorContext (context name)
          506|           (args.${name} or config._module.args.${name})
             |                            ^
          507|       ) (lib.functionArgs f);

```

Can be reproduced by i.e. `nix flake update treefmt-nix` in https://github.com/nix-community/buildbot-nix or most probably any other repo using treefmt-nix and flake.parts

roberth:
Missing `specialArgs` tend to cause infinite recursions.
The flake-parts module doesn't use this entrypoint.
You could add the `specialArgs` to this `submoduleWith` call
https://github.com/numtide/treefmt-nix/blob/3b0afa7961c5d1d94bd02e1d6e560bd84af0525b/flake-module.nix#L35

Alternatively, you could perhaps go through the main entrypoint and get the `type` returned by that `evalModules` call. It should be equivalent to the aforementioned `submoduleWith` call's return value.

JakobLichterfeld (commented):
Nice refactoring, checked the `dart-format` part. lgtm beside of the `flake-parts` issue, see https://github.com/numtide/treefmt-nix/pull/288#issuecomment-2568331807

gabyx:
@zimbatm : Right this change does not change user facing treefmt.nix code?

PR #277, opened 2024-12-10
genemichaels: init

djacu:
Adds a module to use the genemichaels formatter for Rust files that is compatible with version 0.5.7.

PR #289, opened 2024-12-31
feat: update nixpkgs input

djacu:
Brings in genemichaels 0.5.13 (latest).


djacu:
Brings in an update for #277

zimbatm:
thanks. it looks like formatjson5 and elm-format are having some issues.

djacu:
Yeah they are both failing upstream.

https://hydra.nixos.org/build/282571175
https://hydra.nixos.org/build/282877962

I think `formatjson5` has started failing because of the new changes to `missing_docs` in Rust 1.83.
https://releases.rs/docs/1.83.0/
Since the repository has not had any commits in ~1 year,  I think it's best to patch it.

Not sure why `elm-format` has started failing. The hydra logs return a 404 and it only appears to be affecting Darwin. I don't have a capable machine handy so it'll be hard for me to debug. :/

djacu:
I've got a PR in to fix `formatjson5`
https://github.com/NixOS/nixpkgs/pull/370124

Looks like someone else has noticed that `elm-format` is broken as well.
https://github.com/NixOS/nixpkgs/issues/370084

PR #298, opened 2025-01-08
fix: on-unmatched option

brianmcgee:
Also moves excludes in the root of the settings, as global has been deprecated.

Signed-off-by: Brian McGee <brian@bmcgee.ie>

### DIFF
--- a/module-options.nix
+++ b/module-options.nix
@@ -30,25 +30,22 @@ let
     type = types.submodule {
       freeformType = configFormat.type;
       options = {
-        global = {
-          excludes = mkOption {
-            description = "A global list of paths to exclude. Supports glob.";
-            type = types.listOf types.str;
-            default = [ ];
-            example = [ "./node_modules/**" ];
-          };
-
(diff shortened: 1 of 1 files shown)


brianmcgee:
@zimbatm this was never merged. The change still seems relevant when comparing with latest main.

vidhanio:
is there any blockers on this being merged? it is confusing that there exists an option in documentation and in-code that seems to do nothing.

module-options.nix
```diff
@@ -23,32 +23,39 @@ let
...
+  configFile = configFormat.generate "treefmt.toml" settingsData;
+
   # The schema of the treefmt.toml data structure.
   configSchema = mkOption {
     default = { };
     description = "The contents of treefmt.toml";
     type = types.submodule {
       freeformType = configFormat.type;
       options = {
-        global = {
```
vidhanio: should there be an `mkRenamedOptionModule` to notify users to migrate from `global` -> top level settings?

PR #308, opened 2025-02-02
zprint: change args ordering

### LINKED ISSUE #307: mkFormatterModule zprint argument misplacement
Hi @zimbatm ,

After the introduction of mkFormatterModule, I noticed the zprint arguments are misaligned.

The args of "--write" must be placed after zprintOpts.
https://github.com/numtide/treefmt-nix/blob/bebf27d00f7d10ba75332a0541ac43676985dea3/programs/zprint.nix#L16 

Before, options arrange like `[{:width 90} --write]`, now args get put first `[--write {:width 90}]`, which caused errors due to misalignment.

What's the proper fix to make sure args of mkFormatterModule always stay last this case? Thanks.

zimbatm:
Fixes #307


sg-qwt (changes requested):
not mkAfter but mkBefore, since mkBefore is used by default in mkFormatterModule, need something stronger than mkBefore here, for example
`lib.mkOrder 100 (lib.optional (cfg.zprintOpts != null) cfg.zprintOpts);`

sg-qwt (approved):
LGTM

PR #312, opened 2025-02-10
sqlfluff: support sqlfluff lint

m15a:
This PR introduces support for `sqlfluff lint` sub-command in `programs.sqlfluff-lint` module.
Previous `programs.sqlfluff` module has been renamed to `programs.sqlfluff-format` module.

In addition, I've appended command-line options `--disable-progress-bar --processes 0` to both modules' commands for slight performance improvement.


zimbatm (changes requested):
Sounds good but can you keep the formatting with the current name and add `sqlfluff-lint` as the extra program.

m15a:
> keep the formatting with the current name

Thanks for your review. I've reverted to the original name.

PR #325, opened 2025-03-27
sql-formatter: init

xsteadfastx:
Thats one of the formatters im using right now via nvim conform and would love to use in treefmt-nix.


zimbatm:
Nice. Congrats on reaching the 100th formatter!

PR #328, opened 2025-03-28
sql-formatter: fix multi file input

xsteadfastx:
i was a little too fast yesterday. it looks like sql-formatter does not support multi file input. this fixes it.


zimbatm:
no worries. can you also file an upstream issue, so we can eventually remove the fix?

xsteadfastx:
You mean to sql-formatter itself?

Jonas Chevalier ***@***.***> schrieb am Fr., 28. März 2025,
16:17:

programs/sql-formatter.nix
```diff
@@ -52,10 +50,14 @@ in
 
   config = lib.mkIf cfg.enable {
     settings.formatter.sql-formatter = {
-      command = "${cfg.package}/bin/sql-formatter";
-      options = lib.optionals (cfg.dialect != null) [
-        "-l=${cfg.dialect}"
-      ];
+      # sql-formatter doesn't support multiple file targets
```
zimbatm, line 53:
```suggestion
      # sql-formatter doesn't support multiple file targets
      # see https://github.com/sql-formatter-org/sql-formatter/issues/552
```

zimbatm:
thanks!

PR #331, opened 2025-04-02
add oxipng

nim65s:
Hi,

This add https://github.com/shssoichiro/oxipng

Not sure if this counts as a "formatter", but I guess this is still useful :)


programs/oxipng.nix
```diff
@@ -0,0 +1,56 @@
...
+      name = "oxipng";
+      includes = [ "*.png" ];
+    })
+  ];
+
+  options.programs.oxipng = {
+    alpha = lib.mkEnableOption "Perform additional optimization on images with an alpha channel";
+    opt = lib.mkOption {
+      description = "Set the optimization level preset";
+      type = with lib.types; nullOr str;
```
zimbatm, line 24:
```suggestion
      type = with lib.types; str;
```
Since it's not using lib.optionals below. Or use optionals below.

zimbatm (commented):
I like that it's abusing treefmt for something else :)

One comment, the rest should be good to go.

programs/oxipng.nix
nim65s: Yes, there is no need for options here, thanks !

examples/formatter-oxipng.toml
```diff
@@ -0,0 +1,6 @@
+# Example generated by ../examples.sh
+[formatter.oxipng]
+command = "oxipng"
```
Mic92, line 3: Is the compression reproducible?

Mic92: Otherwise it would break CI checks if it tries to re-process files.

PR #333, opened 2025-04-06
fix: potential issue with `cfg.enable` check

famouswizard:
In the original code, there was a conditional check using `lib.mkIf cfg.enable`, which could cause an error if `cfg.enable` was undefined. I’ve updated it to a safer check:  
```nix
lib.mkIf (cfg.enable != null && cfg.enable)
```  
This ensures that `cfg.enable` is not null and is set to true, preventing potential errors when `cfg.enable` is not defined.

### DIFF
--- a/programs/golines.nix
+++ b/programs/golines.nix
@@ -1,9 +1,4 @@
-{
-  lib,
-  config,
-  mkFormatterModule,
-  ...
-}:
+{ lib, config, mkFormatterModule, ... }:
 let
   cfg = config.programs.golines;
 in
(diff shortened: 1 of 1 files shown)


programs/golines.nix
```diff
@@ -25,7 +20,7 @@ in
     })
   ];
 
-  config = lib.mkIf cfg.enable {
+  config = lib.mkIf (cfg.enable != null && cfg.enable) {
```
Mic92, line 23: Is this not of type bool? How can this be null?

Mic92: If something is not set it doesn't have the value null. So this comparison doesn't work.

zimbatm:
thanks, i'm closing this until the issue can be demontrably be reproduced

PR #334, opened 2025-04-06
chore: fix default value for undefined `cfg.enable`

mdqst:
Fixed an issue where `cfg.enable` would not default to `false` if undefined, ensuring the config block is skipped when missing or set to `false`. Also used `lib.mkIf` for proper conditional application.

### DIFF
--- a/programs/fish_indent.nix
+++ b/programs/fish_indent.nix
@@ -13,7 +13,7 @@ in
     })
   ];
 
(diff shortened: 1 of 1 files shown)


zimbatm:
@mdqst do you have a way to reproduce this?

That PR doesn't make sense to me.  In my mental model, `cfg.enable` always exists and defaults to `false` since it's defined by the module system.

That being said, this is the second submission. https://github.com/numtide/treefmt-nix/pull/333 also pointed to the same issue

zimbatm:
It feels like AI

PR #335, opened 2025-04-11
Update nixpkgs input

brianmcgee:
Brings in `treefmt 2.2.1`.


brianmcgee:
`php-cs-fixer` is broke in latest `nixpkgs-unstable`. 

I believe we need to override the `php` version it's using and pin it to `php82` or `php83`. 

How I can do this with the `mkFormatterModule` pattern is unclear.

PR #336, opened 2025-04-13
just: format justfile in subdirectories

Allsimon:
By default, the includes for 'justfile' is limited to the root directory. 

This change introduces a new behavior where  'justfile' files are now matched across subdirectories.


Allsimon:
After reading a bit more how treefmt works, I realize this is a bad approach

I'll re-open this with an enhanced version

PR #337, opened 2025-04-15
feat: Add buf as formatter

sebimarkgraf:
Adds the [buf formatter](https://github.com/bufbuild/buf) for protobuf files.

Performed test on the example from here the [protobuf docs](https://protobuf.dev/).


w4:
Thanks for this, looks like `buf format` only accepts at most one file though

PR #338, opened 2025-04-15
programs.dprint: update plugins example

phanirithvij:
depends on https://github.com/NixOS/nixpkgs/pull/398815

I want to simplify this example ideally.

cc @zimbatm

### DIFF
--- a/programs/dprint.nix
+++ b/programs/dprint.nix
@@ -8,6 +8,7 @@
 let
   inherit (lib)
     filterAttrsRecursive
+    literalExpression
     mkIf
     mkOption
     optionals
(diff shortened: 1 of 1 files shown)


phanirithvij:
did I do something to buildbot? its in a loop.

sorry :sob:

phanirithvij:
@Mic92 it depends on the linked nixpkgs pr. But for now it won't break anything because it is just the example text, but best to revert it.

Or you can also review the linked nixpkgs pr.

Mic92:
Reverted in 8d404a69efe76146368885110f29a2ca3700bee6

Issue #247, opened 2024-10-12
Clarification on how to configure mypy


Ma27:
**Describe the bug**

With the `mypy` check from this repo, no files get checked with mypy on a treefmt run.

**To Reproduce**

1. Given `foo/bar.py`
2. Given the following `treefmt-nix` configuration (in `flake.nix`):
```nix
{
  treefmt = {
    projectRootFile = "flake.nix";
    programs = {
      nixfmt.enable = true;
      ruff.enable = true;
      mypy = {
        enable = true;
        directories = {
          ".".modules = [ "foo" ];
        };
      };
    };
  };

}
```
4. Now running `treefmt` doesn't give any `mypy` errors.

I think part of the problem is the following in the final treefmt config:

```
includes = ["./foo/**/*.py"]
```

This doesn't seem to give any matches in treefmt.

After that I replaced the `".".modules = ["foo"]` with `".".files = [ "*.py" ]`. With that I got the following error:

```
+ /nix/store/ryp4vp6rs3d1gk0mgj6x8bncavsp4wqa-python3.12-mypy-1.10.1/bin/mypy '' '*.py'
mypy: can't read file '*.py': No such file or directory
```

This is because `*.py` isn't a recursive glob, however it's the glob needed for `treefmt` to validate a change in any Python file in a repo with `mypy`.

**Expected behavior**

I don't know, not sure if this is even a bug.

What am I supposed to do to let mypy run against a given module (`foo`)?

**System information**

NixOS 24.05, treefmt 2.0.5 (via treefmt-nix 879b29ae9a0378904fbbefe0dadaed43c8905754), git-hooks.nix at rev 85f7a7177c678de68224af3402ab8ee1bcee25c8.

**Additional context**

declension:
FWIW, we long ago gave up on `treefmt` for `mypy` and use custom derivations (or checks) instead. But it would be great (for caching, especially) if treefmt could support mypy properly.

zimbatm:
It seems to be something specific to how mypi works.

Other people that contributed to the module: @Mic92 and @DavHau

Mic92:
This is how I use mypy: https://git.clan.lol/clan/clan-core/src/commit/446ebb8e336b60a5f2a27f21742d93d4cfa54f0d/formatter.nix#L97

Mic92:
Looks like the globbing changed in treefmt? Before it '/**/*.py' used to match files, but this is no longer the case. @brianmcgee is this no longer supported syntax?

[...]

PR #340, opened 2025-04-16
mypy: fix recursive globbing

Mic92:
fixes https://github.com/numtide/treefmt-nix/issues/247

### DIFF
--- a/programs/mypy.nix
+++ b/programs/mypy.nix
@@ -55,12 +55,6 @@ in
                 ];
                 description = "Modules to check";
               };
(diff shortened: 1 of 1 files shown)


Mic92:
fixes https://github.com/numtide/treefmt-nix/issues/247

programs/mypy.nix
```diff
@@ -58,7 +58,7 @@ in
               files = lib.mkOption {
```
Mic92, line 58: Not quite sure to support this use case actually in conjunction with modules...

PR #341, opened 2025-04-18
programs.dprint: update plugins example

phanirithvij:
depends on https://github.com/NixOS/nixpkgs/pull/398815

### DIFF
--- a/programs/dprint.nix
+++ b/programs/dprint.nix
@@ -8,6 +8,7 @@
 let
   inherit (lib)
     filterAttrsRecursive
+    literalExpression
     mkIf
     mkOption
     optionals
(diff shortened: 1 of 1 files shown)


phanirithvij:
@Mic92 I think it makes sense to merge this only after updating nixpkgs in this repo's flake?

Mic92:
I'd say we wait for https://nixpk.gs/pr-tracker.html?pr=341
It won't work in nixpkgs stable yet, but I don't think this is a big deal because we are not regressing on existing behavior.

phanirithvij:
https://nixpk.gs/pr-tracker.html?pr=398815

PR #346, opened 2025-04-27
just: format justfile in subdirectories

Allsimon:
Given the following projects:
```
.
├── foobar
│   └── justfile
└── justfile
```
the previous implementation would have just reformatted `./justfile` but not `./foobar/justfile`.

I also simplified the previous bash implementation by re-using mkFormatterModule

### DIFF
--- a/programs/just.nix
+++ b/programs/just.nix
@@ -6,6 +6,8 @@
   ...
 }:
 let
+  # Search 'justfile' in subdirectories too. This isn't needed for filename starting with `*.`.
+  alsoSearchSubdirectory = lib.lists.concatMap (fileName: [ fileName ] ++ [ "*/${fileName}" ]);
   cfg = config.programs.just;
 in
 {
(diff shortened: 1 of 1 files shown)


programs/just.nix
```diff
@@ -1,40 +1,32 @@
...
       name = "just";
-      includes = [
-        "[Jj][Uu][Ss][Tt][Ff][Ii][Ll][Ee]" # 'justfile', case insensitive
-        ".[Jj][Uu][Ss][Tt][Ff][Ii][Ll][Ee]" # '.justfile', case insensitive
-        "*.[Jj][Uu][Ss][Tt]" # '.just' module, case insensitive
-        "*.[Jj][Uu][Ss][Tt][Ff][Ii][Ll][Ee]" # '.justfile' module, case insensitive
+      args = [
+        "--fmt"
+        "--unstable"
+        "--justfile"
```
katexochen, line 19: This simplification is wrong. just itself doesn't comply with the treefmt spec, as the `--justfile` flags expects a single argument, but the spec requires the formatter to accept multiple file names as arguments (see https://treefmt.com/latest/reference/formatter-spec/#1-files-passed-as-arguments). That's the reason for the bash loop (I should have added a comment, sorry for that).

katexochen (changes requested):

programs/just.nix
Allsimon: Thanks, I didn't notice that at all. I re-added the bash loop :)

PR #350, opened 2025-04-30
fix defaultText for `pkgs` option using `literalMD`

MattSturgeon:
Passing a non-literal string to `example` or `defaultText` causes the docs to render the value using `lib.generators.toPretty`.

If you want to use markdown, use `literalMD`. If you want to pass a nix expression, use `literalExpression`.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -38,7 +38,7 @@ in
               {
                 options.pkgs = lib.mkOption {
                   default = pkgs;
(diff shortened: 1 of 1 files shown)


zimbatm:
thanks! do you mind rebasing the change?

MattSturgeon:
> thanks! do you mind rebasing the change?

Done :+1:

zimbatm:
thanks!

PR #356, opened 2025-05-16
Wrap `buf` correctly, as it does not take multiple file targets

drew-dirac:
`buf` does not accept multiple file targets. Because of this, setting `programs.buf.enable = true;` resulted in the following error:

```
ERRO formatter | buf: failed to apply with options '[format -w]': exit status 1

Usage:
  buf format <source> [flags]

Flags:
      --config string          The buf.yaml file or data to use for configuration
  -d, --diff                   Display diffs instead of rewriting files
      --disable-symlinks       Do not follow symlinks when reading sources or configuration from the local filesystem
                               By default, symlinks are followed in this CLI, but never followed on the Buf Schema Registry
      --error-format string    The format for build errors printed to stderr. Must be one of [text,json,msvs,junit,github-actions] (default "text")
      --exclude-path strings   Exclude specific files or directories, e.g. "proto/a/a.proto", "proto/a"
                               If specified multiple times, the union is taken
      --exit-code              Exit with a non-zero exit code if files were not already formatted
  -h, --help                   help for format
  -o, --output string          The output location for the formatted files. Must be one of format [dir,protofile]. If omitted, the result is written to stdout (default "-")
      --path strings           Limit to specific files or directories, e.g. "proto/a/a.proto", "proto/a"
                               If specified multiple times, the union is taken
  -w, --write                  Rewrite files in-place

Global Flags:
      --debug               Turn on debug logging
      --log-format string   The log format [text,color,json] (default "color")
      --timeout duration    The duration until timing out, setting it to zero means no timeout (default 2m0s)

accepts at most 1 arg(s), received 59
```

This PR wraps `buf` in a similar way to https://github.com/numtide/treefmt-nix/blob/main/programs/statix.nix to address this issue.


zimbatm:
Thanks! I pushed some small fixed and once the CI pass it should be good to go.

/cc @sebimarkgraf who submitted the original PR

PR #359, opened 2025-05-20
Add support for rstfmt

rbpatt2019:
Adds support for "rstfmt is a tool for automatically formatting reStructuredText files in a
consistent way."

I couldn't see any formatters for reStructuredText supported by treefmt-nix, so I added one. The only option I have changed from default is line length, which I have set to 80, rather than the default of 72.


zimbatm:
thanks!

PR #361, opened 2025-05-25
Add cabal-gild, a cabal formatter

jian-lin:


README.md
```diff
@@ -221,7 +221,7 @@ functions.
 <!-- `> bash ./supported-programs.sh` -->
 
 <!-- BEGIN mdsh -->
-`treefmt-nix` currently supports 106 formatters:
+`treefmt-nix` currently supports 108 formatters:
```
jian-lin, line 224: The number should be correct if this PR is merged after https://github.com/numtide/treefmt-nix/pull/360.

programs/cabal-gild.nix
```diff
@@ -0,0 +1,43 @@
...
+  config = lib.mkIf cfg.enable {
+    settings.formatter.cabal-gild = {
+      # cabal-gild doesn't support multiple file targets, see upstream issue 35
+      command = pkgs.writeShellApplication {
+        name = "cabal-gild-wrapper";
+        text = ''
+          for file in "$@"; do
+            ${lib.getExe cfg.package} --io="$file"
+          done
+        '';
```
jian-lin, line 39: `args` / `options` is not supported.  Should we fix that?  How?

programs/cabal-gild.nix
```diff
@@ -0,0 +1,43 @@
...
+        "*.cabal"
+        "*cabal.project"
+        "*cabal.project.local"
+      ];
+    })
+  ];
+
+  config = lib.mkIf cfg.enable {
+    settings.formatter.cabal-gild = {
+      # cabal-gild doesn't support multiple file targets, see upstream issue 35
```
jfly, line 32: nit: Please include a direct link here.

programs/cabal-gild.nix
```diff
@@ -0,0 +1,43 @@
...
+      command = pkgs.writeShellApplication {
+        name = "cabal-gild-wrapper";
+        text = ''
+          for file in "$@"; do
+            ${lib.getExe cfg.package} --io="$file"
```
jfly, line 37: Are you sure we don't need `--input="$file"` here? From https://hackage.haskell.org/package/cabal-gild:

> `--input=FILE`: Uses `FILE` as the input. If this is `-` (which is the default), then the input will be read from `STDIN`.

programs/cabal-gild.nix
jfly: I'm not sure I understand the question. I don't think tools in treefmt-nix *have* to have args or options.

jian-lin: If users set `programs.cabal-gild.args = ...`, probably they expect those args to take effect.  In the current implementation, those args are ignored.

Do you mean ignoring args set by users is OK?

programs/cabal-gild.nix
jian-lin: Yes, I am sure and it works well in my projects.

I set `--io`and here is the doc:

--io=FILE: Shortcut for setting both --input=FILE and --output=FILE at the same time. This is useful for formatting a file in-place.

examples/formatter-cabal-gild.toml
```diff
@@ -0,0 +1,6 @@
+# Example generated by ../examples.sh
+[formatter.cabal-gild]
+command = "cabal-gild"
```
jian-lin, line 3: How should I fix this [CI error][1]?

Change `command` in this example to `cabal-gild-wrapper` or change the name of the wrapper script to `cabal-gild`?  Both of them seem not correct to me.

[1]: https://buildbot.numtide.com/#/builders/312/builds/367

programs/cabal-gild.nix
jfly: Ohhhh sorry now I understand.

Is this a problem with all the `for file in "$@"` wrappers we have in treefmt-nix?

examples/formatter-cabal-gild.toml
jfly: I guess just change it to `cabal-gild-wrapper`. Ideally we'd have a note here about how these tools need a wrapper to work. Note how this is broken with other tools that need similar `for file in ` wrapper:

- https://github.com/numtide/treefmt-nix/blob/020cb423808365fa3f10ff4cb8c0a25df35065a3/examples/formatter-statix.toml
- https://github.com/numtide/treefmt-nix/blob/020cb423808365fa3f10ff4cb8c0a25df35065a3/examples/formatter-buf.toml

programs/cabal-gild.nix
jian-lin: Yes.  `$@` contains `args`/`options`.  An example value of `$@` can be `arg1 arg2 file1 file2`.

jfly: What you've done here looks the same as `buf.nix`, `statix.nix`, and others. I think it's fine.

jfly (approved):

jian-lin:
I think this PR is ready.  Hope it can be merged.

`cabal-gild` is the default formatter of haskell language server.  So I think this PR will benefit many people.

Issue #357, opened 2025-05-17
Issue writing golden test in terms of treefmt-nix.


locallycompact:
I have rewritten a simple golden test in terms of treefmt-nix here:

https://gitlab.horizon-haskell.net/nix/gitlab-ci/-/merge_requests/8

I have added a newline to the golden file to trigger it to fail, and it fails as expected when running the treefmt check via nix flake check, however running the formatter does not reformat the golden file.

Reproduction:

```
git clone https://gitlab.horizon-haskell.net/nix/gitlab-ci
cd gitlab-ci
git checkout rewrite-golden-test
cd test/golden-simple
nix flake check # fails as expected
nix run .#formatter.x86_64-linux # does not reformat the golden file
```

Did I misconfigure something?

locallycompact:
So it seems like the check is traversing just the 8 files in the test flake folder, but the formatter is traversing the entire repository of 32 files

```
lc@earthrealm ~/S/g/n/g/t/golden-simple (rewrite-golden-test) [1]> nix flake check -L
warning: Git tree '/home/lc/Source/gitlab.horizon-haskell.net/nix/gitlab-ci' is dirty
treefmt-check> treefmt v2.2.1traversed 8 files
treefmt-check> emitted 1 files for processing
treefmt-check> formatted 1 files (1 changed) in 19ms
treefmt-check>  M modules/golden
treefmt-check> diff --git a/modules/golden b/modules/golden
treefmt-check> index 44d0f0a..644cd81 100644
treefmt-check> --- a/modules/golden
treefmt-check> +++ b/modules/golden
treefmt-check> @@ -11,5 +11,3 @@ test:
treefmt-check>    - echo "Success"
treefmt-check>    stage: test
```

```
lc@earthrealm ~/S/g/n/g/t/golden-simple (rewrite-golden-test)> nix fmt
warning: Git tree '/home/lc/Source/gitlab.horizon-haskell.net/nix/gitlab-ci' is dirty
traversed 32 files
emitted 0 files for processing
formatted 0 files (0 changed) in 22ms
```

and therefore the formatter is not finding the golden file, but why do these behave differently?

jfly:
[...]

PR #364, opened 2025-06-03
Use `flake.nix` for the project root when dealing with flakes

jfly:
The default value for `config.projectRootFile` is `.git/config`, which is a nice default that doesn't assume flakes (because flakes are more controversial than git in the nix community :))

However, if you're using flake-parts, then you're definitely OK with flakes, and using `flake.nix` as a project root file makes sense.

As a consequence of this, subflakes now probably do what you expect. In other words, this fixes https://github.com/numtide/treefmt-nix/issues/357.

Side note: in the future, nix will expose a `PRJ_ROOT` env var that we could use instead. See
https://github.com/NixOS/nix/commit/76358748da95528f1a665455dae35dad61081dcb.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -62,6 +62,8 @@ in
                     Path to the root of the project on which treefmt operates
                   '';
                 };
+
+                config.projectRootFile = lib.mkDefault "flake.nix";
               }
             ];
           };


zimbatm:
thanks :)

PR #383, opened 2025-07-08
Add `formatter.useEditorconfig` to Biome

oskardotglobal:
Title. The added option makes Biome respect editorconfig.
See https://biomejs.dev/reference/configuration/#formatteruseeditorconfig

### DIFF
--- a/programs/biome.nix
+++ b/programs/biome.nix
@@ -138,6 +138,13 @@ in
             example = true;
             default = false;
           };
+
+          useEditorconfig = l.mkOption {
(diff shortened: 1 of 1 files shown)


jfly (approved):

oskardotglobal:
Sorry, I tested this further and it doesn't seem to be working, gonna make this a draft for now

hackardoX:
Hi @oskardotglobal, now the whole biome config is supported. However, this could still not working as expected (my guess is that biome is looking for the .editorconfig in the nix store instead of the active workspace)

oskardotglobal:
Yeah, seems like a problem with biome

PR #384, opened 2025-07-10
fix(cabal): cabal-fmt and cabal-gild discover module

ncaq:
`cabal-fmt` and `cabal-gild` can discover module and autocomplete.
```cabal
-- cabal-gild: discover ./src
exposed-modules:
```
The current system, treefmt can't detect change haskell module.
I write to depend to haskell module file.


ncaq:
Maybe I mistake.
cabal-gild it's don't work.

ncaq:
I tested my product and cabal-gild.
My way code, correct way in this project?

PR #391, opened 2025-07-28
feat: use json schema for biome

andrea11:
I'm writing this PR because I noticed that the formatter is missing most of the options that Biome provides. I think it would be a lot easier for us if we didn't have to manually add our own options. Instead, I propose that we use free-field options and validate them against Biome's JSON schema.
I have also provided the option to set the command for biome and if it should process unsafe fixes.
I am not an expert of Nix, I tested locally and it is working correctly, but please let me know if any improvement is required

### DIFF
--- a/programs/biome.nix
+++ b/programs/biome.nix
@@ -13,6 +13,11 @@ let
   cfg = config.programs.biome;
   biomeVersion = if builtins.match "^1\\." pkgs.biome.version != null then "1.9.4" else "2.1.2";
   schemaUrl = "https://biomejs.dev/schemas/${biomeVersion}/schema.json";
+  schemaSha256s = {
+    "1.9.4" = "sha256:0yzw4vymwpa7akyq45v7kkb9gp0szs6zfm525zx2vh1d80568dlz";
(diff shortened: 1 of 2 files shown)


hackardoX:
@zimbatm @jfly What do you think about this PR?

programs/biome.nix
```diff
@@ -11,6 +10,10 @@ let
   t = l.types;
   p = pkgs;
 
+  cfg = config.programs.biome;
+  biomeVersion = if builtins.match "^1\\." pkgs.biome.version != null then "1.9.4" else "2.1.2";
```
jfly, line 14: A couple thoughts:

1. I am not really understanding this ternary. How does 1.x equate to 1.9.4, and how does "not 1.x" equate to 2.1.2? This either needs reworking, or at least a comment.
2. The `if biomeVersion == "1.9.4" then` down below should get moved closer to this code. It feels very related, but is so far away it'll be more likely that things will get out of whack.

sportshead:
LGTM, definitely a better solution than before

zimbatm:
thanks!

PR #396, opened 2025-08-06
chore: update nixpkgs

zimbatm:

PR #398, opened 2025-08-06
chore: format with updated `nixfmt`

a-kenji:
- **feat: Update `nixpkgs` input**
- **chore: Format with updated `nixfmt`**


a-kenji:
Will be handled in #396

PR #405, opened 2025-08-24
feat: auto update biome version

hackardoX:
The purpose of this PR is to automatically create PR to keep biome hashes updated. 
### What this PR does:

- Biome now store all hashes for the related schema, so it will be compatible if whatever nixpkgs user has
- New script to automatically retrieve newest versions of biome schema, and automatically update [https://github.com/numtide/treefmt-nix/blob/main/programs/biome.nix](biome.nix)
- GitHub action running every day to check if a new biome is available and create a PR if so (it takes around 1 minute: [Test build](https://github.com/hackardoX/treefmt-nix/actions/runs/17195210282) ). Resulting in a [PR created by GitHub Bot](https://github.com/hackardoX/treefmt-nix/pull/1) 

### Tests

- [x] Tested locally in other flakes

### DIFF
--- a/.github/workflows/update-biome.yml
+++ b/.github/workflows/update-biome.yml
@@ -1,0 +1,34 @@
+name: Update Biome Schemas
+on:
+  schedule:
+    - cron: "0 12 * * *" # Runs at 12:00 UTC every day
+  workflow_dispatch: # allow manual triggering
+jobs:
+  update-biome:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout repository
+      - name: Checkout repo
+        uses: actions/checkout@v5
+      # Setup nix
+      - name: Setup Nix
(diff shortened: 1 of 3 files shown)


hackardoX:
Hi @zimbatm, what do you think about this approach? I am not sure if you are ok to integrate GitHub action in the repository. Let me know

hackardoX:
Hi @zimbatm @jfly, could you have a look to this PR?

zimbatm:
thanks, I think this is out of scope too. treefmt-nix relies on nixpkgs for packaging.

hackardoX:
Thanks for the feedback @jfly @zimbatm. Currently biome does not have any command to validate the config, maybe it can be requested. Doing so, treefmt-nix could accept any type of properties for biome config, and use the command to validate (if needed) at build time.

I would close this PR.

PR #407, opened 2025-09-17
nixfmt-classic: remove warning

Enzime:
I found it annoying to always have this warning especially as I'm already explicitly opting to use this formatter and I thought that it would just add extra clutter to have an option to disable the warning

### DIFF
--- a/programs/nixfmt-classic.nix
+++ b/programs/nixfmt-classic.nix
@@ -1,12 +1,4 @@
-{
-  lib,
-  config,
-  mkFormatterModule,
-  ...
-}:
-let
-  cfg = config.programs.nixfmt-classic;
-in
+{ mkFormatterModule, ... }:
 {
   meta.maintainers = [ ];
 
@@ -16,16 +8,4 @@ in
       includes = [ "*.nix" ];
     })
   ];
-
-  config = lib.mkIf cfg.enable {
-    settings.formatter.nixfmt-classic = (
-      lib.warn ''
-        nixfmt-classic is the original flavor of 'nixfmt'.
-        This version differs considerably from the new standard and is currently
-        unmaintained.
-        It has been superseded by 'nixfmt', which conforms to the
-        'nixfmt-rfc-style' standard.
-      '' { }
-    );
-  };
 }


jfly (commented):
I don't think this is a good idea. Anyone using `nixfmt-classic` should move off of it, or find an out-of-tree solution. We intend to drop nixfmt-classic from nixpkgs ([timing TBD](https://github.com/NixOS/nixfmt/issues/340)).

MattSturgeon:
I personally feel like nixpkgs is the correct place for such a warning, not treefmt-nix. So I agree with the spirit of this PR.

However, in lieu of such a warning actually being added to nixpkgs, I don't see any reason for treefmt-nix to rush into dropping its warning; assuming we actually do indeed intend to deprecate and/or drop nixfmt-classic from nixpkgs in the foreseeable future.

My preference would be to first decide in https://github.com/NixOS/nixfmt/issues/340 if/when `nixfmt-classic` will be deprecated, then we will know how treefmt-nix should handle its warnings.

Either the warning will be dropped because there is no plan to deprecate, or the warning will be dropped because it is moved to nixpkgs.

programs/nixfmt-classic.nix
```diff
@@ -18,14 +18,6 @@ in
...
-      lib.warn ''
-        nixfmt-classic is the original flavor of 'nixfmt'.
-        This version differs considerably from the new standard and is currently
-        unmaintained.
-        It has been superseded by 'nixfmt', which conforms to the
-        'nixfmt-rfc-style' standard.
-      '' { }
-    );
+    settings.formatter.nixfmt-classic = { };
   };
```
zimbatm, line 22:
```suggestion
```

The whole block can be dropped

zimbatm (approved):
Make sense to me. People opting for nixfmt-classic have made a conscious choice and don't need to be reminded about it at every eval. The warning is an overhang when we switched the "nixfmt" formatter around.

Enzime:
Due to the Nix ecosystem generally not having disable-able warnings, I would probably caution against any warnings regarding formatters as it just adds to the noise and formatters are tools that can be easily swapped between so a migration doesn't need to be done before a tool is completely EOL/removed.

PR #414, opened 2025-09-29
sqlfluff: Add option for configFile

matthiasbeyer:
I confirmed that this works as expected.

If someone tells me how to also add a configuration option that can be used from within nix without mapping all [config options sqlfluff has](https://docs.sqlfluff.com/en/stable/configuration/default_configuration.html) to nix options by hand, I'd love to do that.

For now, this works for me.

### DIFF
--- a/programs/sqlfluff.nix
+++ b/programs/sqlfluff.nix
@@ -46,6 +46,10 @@ in
         "--disable-progress-bar"
         "--processes"
         "0"
+      ]
+      ++ lib.optionals (cfg.configFile != null) [
+        "--config"
+        "${cfg.configFile}"
       ];
       includes = [ "*.sql" ];
     })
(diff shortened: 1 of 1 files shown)


shimeoki:
> If someone tells me how to also add a configuration option that can be used from within nix without mapping all [config options sqlfluff has](https://docs.sqlfluff.com/en/stable/configuration/default_configuration.html) to nix options by hand, I'd love to do that.

you mean like "using one of the options from the sqlfluff configuration file, but don't pass it as a config file"?

i think just using the config is the best choice here. it's probably out of scope for treefmt (and treefmt-nix): it's a multiplexer for using multiple formatters at once to configure their behaviour from this standpoint. formatter cli options via treefmt is ok, but trying to translate even the configuration file feels kinda off.

if you want to just configure the options from nix to not use toml (if i am not wrong that this config's format is toml) by hand, then you can make a nix attrset and build a toml derivation from it (via `pkgs.formats`? i am not sure), and use it as a config file. i don't know if it's worth it, though

PR #415, opened 2025-09-30
feat(shfmt): add simplify option

shunueda:
Adding support for `-s` (`--simplify`) flag in shfmt.

`true` by default to avoid breaking change.

https://man.archlinux.org/man/extra/shfmt/shfmt.1.en#Parser_flags


shunueda:
@zimbatm could I get a review on this? Thank you in advance!

zimbatm:
LGTM, thanks!

PR #420, opened 2025-10-16
djlint: init

tfc:
Add the `djlint` tool for django, jinja, html formatting/linting


tfc:
@zimbatm thank you for the quick reaction!

PR #422, opened 2025-10-19
feat: allow customizing packages with plugins, add such customization for mdformat

acid-bong:
This will allow customizing formatters, that can be bundled with
extentions/plugins (like `mdformat.withPlugins (p: [...])`), with
module options. Example:
```nix
# definition
{
  options.programs.mdformat = {
    plugins = lib.mkOption {
      type = with lib.types; functionTo (listOf package);
      default = _: [ ];
    };
  };
  config.programs.mdformat.finalPackage = cfg.package.withPlugins cfg.plugins;
}

# usage
{
  programs.mdformat = {
    enable = true;
    plugins = ps: [
      mdformat-footnote
      mdformat-gfm
    ];
  };
}
```

Inspired by https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/x11/window-managers/qtile.nix and some others.

---

i hope i'm not overengineering this

### DIFF
--- a/programs/mdformat.nix
+++ b/programs/mdformat.nix
@@ -22,57 +22,84 @@ in
     on the project's GitHub page:
     <https://github.com/hukkin/mdformat>
   */
-  options.programs.mdformat.settings =
+  options.programs.mdformat =
     let
       inherit (lib.types)
         bool
         enum
+        functionTo
         int
+        listOf
         nullOr
         oneOf
+        package
         ;
     in
     {
-      end-of-line = lib.mkOption {
+      plugins = lib.mkOption {
         description = ''
-          Output file line ending mode.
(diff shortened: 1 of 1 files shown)


programs/mdformat.nix
```diff
@@ -13,10 +13,6 @@ in
   imports = [
     (mkFormatterModule {
       name = "mdformat";
-      package = [
-        "python3Packages"
-        "mdformat"
-      ];
```
zimbatm: I didn't understand this change. Is the package defined somewhere else?

acid-bong:
> one or two might benefit from that same treatment

Prettier has an unused `plugins` option, although it's of `listOf str` type (and thus doesn't fit here). I guess i'll have a glance

zimbatm:
thanks!

PR #429, opened 2025-11-04
Add in config file option for clang-format

rosscarter154:
This PR adds in a config file option for clang format to allow for clang-format configurations to be used from a nix store path.

### DIFF
--- a/programs/clang-format.nix
+++ b/programs/clang-format.nix
@@ -1,13 +1,27 @@
-{ mkFormatterModule, ... }:
+{
+  mkFormatterModule,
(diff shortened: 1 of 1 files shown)


rosscarter154:
> Is this a normal pattern in treefmt-nix? I thought we would instead create a rfc42 style freeform option instead?

I'm not familiar with what you mean by that. Could you maybe show an example of how I would use that with the existing definition of the clang-format module

rosscarter154:
I was able to do this just by overriding the formatter options

PR #436, opened 2025-11-17
fix(various): don't include empty config

acid-bong:
Some formatters generate config files with default/no settings. This change will allow using an in-tree config file with those formatters as well (for example, when one wants identical formatting in Treefmt and their editor)


acid-bong:
i see multiple formatters have that issue, will come back with a larger diff

acid-bong:
actually no, i'll try adapt Gene Michaels too

PR #443, opened 2025-12-03
feat(shfmt): add a `useEditorConfig` option

jfly:
With vanilla `shfmt`, it's easy to get it to use `.editorconfig` files: you simply [invoke it with no
arguments](https://github.com/mvdan/sh/blob/v3.12.0/cmd/shfmt/shfmt.1.scd?plain=1#L20-L21).

However, with `treefmt-nix`, it's a bit more challenging because we turn on some knobs by default, and those knobs change over time. For example, we have users who [used to accomplish this by setting `indent_size` to `null`](https://github.com/sellout/bash-strict-mode/blob/37eaa2e33d24168e0a903ddd9828666c1b91f867/.config/project/default.nix#L31-L33), but then we recently added the [simplify flag which defaults to true](https://github.com/numtide/treefmt-nix/commit/6b850bab259649cfcb25a54280ca4c0b42a40358), thereby causing shfmt to no longer honor `.editorconfig` files.

To give people some stability, this new `useEditorConfig` option changes the defaults so they're all disabled, and furthermore asserts that every option is disabled (in case people try to change them by hand). This protects our users against new default options in the future.

An alternative to this would be a policy of not deviating from shfmt's defaults, but I think this is clearer for everyone.

### DIFF
--- a/programs/shfmt.nix
+++ b/programs/shfmt.nix
@@ -31,30 +31,40 @@ in
   options.programs.shfmt = {
     indent_size = lib.mkOption {
       type = lib.types.nullOr lib.types.int;
-      default = 2;
+      default = if cfg.useEditorConfig then null else 2;
       example = 4;
       description = ''
         Sets the number of spaces to be used in indentation. Uses tabs if set to
-        zero. If this is null, then [.editorconfig will be used to configure
-        shfmt](https://github.com/patrickvane/shfmt#description).
+        zero.
       '';
     };
 
     simplify = lib.mkOption {
       type = lib.types.bool;
-      default = true;
+      default = !cfg.useEditorConfig;
       description = ''
         Enables the `-s` (`--simplify`) flag, which simplifies code where possible.
       '';
     };
+
+    useEditorConfig = lib.mkEnableOption ''
+      Use .editorconfig file for
+      formatting. This is mutually exlusive with all other settings. See [shfmt
+      docs](https://github.com/mvdan/sh/blob/v3.12.0/cmd/shfmt/shfmt.1.scd?plain=1#L20-L21).
+    '';
   };
 
   config = lib.mkIf cfg.enable {
     settings.formatter.shfmt.options =
-      (lib.optionals (!isNull cfg.indent_size) [
-        "-i"
-        (toString cfg.indent_size)
-      ])
-      ++ (lib.optionals (cfg.simplify) [ "-s" ]);
+      let
+        options =
+          (lib.optionals (!isNull cfg.indent_size) [
+            "-i"
+            (toString cfg.indent_size)
+          ])
+          ++ (lib.optionals (cfg.simplify) [ "-s" ]);
+      in
+      assert cfg.useEditorConfig -> builtins.length options == 0;
+      options;
   };
 }


l0b0 (approved):
This is much clearer, thank you!

jfly:
Pinging the maintainers of this formatter. @zimbatm, @katexochen, could I ask for a review?

sellout:
I haven’t tried this PR yet, but I ran into this problem, and I don’t _think_ the `-s` change (#415) should have affected it. [shfmt’s README](https://github.com/patrickvane/shfmt#readme) claims “If any parser or printer flags are given to the tool, no EditorConfig files will be used,” but

1. `-s` isn’t listed as either a parser or printer flag and
2. treefmt-nix hardcoded `-s` before, when setting `shfmt.indent_size = null` worked, that PR only made it optional, it didn’t add it.
3. setting `shfmt.simplify = false` isn’t fixing the issue for me.

But I also don’t know what else may have triggered it.

sellout:
Hrmm, I must have done something wrong the first time around – it looks like `shfmt.simplify = false` _does_ address my issue, but the other two points stand, so maybe it was a change in shfmt that caused the new behavior?

In any case, I think `useEditorConfig` is a good idea.

jfly:
> [shfmt’s README](https://github.com/patrickvane/shfmt#readme) claims “If any parser or printer flags are given to the tool, no EditorConfig files will be used,” but

For the record (as this confused me for a while), you're linking to the wrong project here. It's a fork of mvdan, which *is* packaged by nixpkgs. You'll see I fixed the link in this PR.

> `-s` isn’t listed as either a parser or printer

This sounds like a docs issue with shfmt. The code clearly disables editorconfig if simplify is enabled, see https://github.com/mvdan/sh/blob/8202166b7d1e3473a7c65eeac53ddbdb55d5b808/cmd/shfmt/main.go#L235

> treefmt-nix hardcoded -s before, when setting shfmt.indent_size = null worked, that PR only made it optional, it didn’t add it.

Are we looking at the same PR? In https://github.com/numtide/treefmt-nix/pull/415/changes, I see no evidence that we ever hardcoded `-s`, and that PR *did* make the new simplify default to enabled.

sellout:
**NB**: Just responding to a comment, I think this PR is correct as-is and I’m glad it’s merged.

> > treefmt-nix hardcoded -s before, when setting shfmt.indent_size = null worked, that PR only made it optional, it didn’t add it.
> 
> Are we looking at the same PR? In https://github.com/numtide/treefmt-nix/pull/415/changes, I see no evidence that we ever hardcoded `-s`, and that PR _did_ make the new simplify default to enabled.

A few things led me to that conclusion:

1. the PR description says “`true` by default to avoid breaking change” (implying that `-s` had previously been set);
2. [the ‘examples/formatter-shfmt.toml‘ diff](https://github.com/numtide/treefmt-nix/pull/415/files#diff-b7b35dc56409f8176a31a38f8188b2606a060e315d5493d0aad2fd58e241e746L6-R6) showed that `-s` was set before this change; and
3. [the call to `mkFormatterModule`](https://github.com/numtide/treefmt-nix/pull/415/files#diff-e2e8efa1aaf8811b132dd8907ad7988353c2d67c0571fd73a10457ff78143b6dL20-R20) previously passed `-s`.

I still don’t think #415 is to blame (primarily because of no. 3).

Digging in, [mvdan/sh 3.12.0 is the first release to treat `-s` as a parser flag](https://github.com/mvdan/sh/commit/8c1554a9c2737ec9e04a57e3db670867426627f9), so it looks like [the upgrade of shfmt in Nixpkgs 25.11](https://github.com/NixOS/nixpkgs/commit/3cdf44d1c4ada594a942eb72cad5166193f1d5bb) is the cause of the breakage that this fixes.

-----

Incidentally,

> > `-s` isn’t listed as either a parser or printer
> 
> This sounds like a docs issue with shfmt.

The correct repo _does_ list it as a parser flag (as of 3.12.0) – it’s only the fork that doesn’t (which is probably correct for that fork).

jfly:
> Digging in, [mvdan/sh 3.12.0 is the first release to treat -s as a parser flag](https://github.com/mvdan/sh/commit/8c1554a9c2737ec9e04a57e3db670867426627f9), so it looks like [the upgrade of shfmt in Nixpkgs 25.11](https://github.com/NixOS/nixpkgs/commit/3cdf44d1c4ada594a942eb72cad5166193f1d5bb) is the cause of the breakage that this fixes.

My bad, thanks for clearing this up!

PR #445, opened 2025-12-07
feat(beautysh): include .bash extension and Direnv files

acid-bong:
In consistency with presets for Shfmt and Shellcheck


acid-bong:
bloody examples

PR #451, opened 2025-12-23
fix(dprint): fix file include pattern to include all files

mputz86:
Instead of only hidden files. The problem was probably introduced while migrating to `mkFormatterModule`.

Currenty, when using the dpint program, it fails with:
```
No files found to format with the specified plugins at /home/(removed). You may want to try using `dprint output-file-paths` to see which files it's finding or run with `--allow-no-files`.
```

### DIFF
--- a/examples/formatter-dprint.toml
+++ b/examples/formatter-dprint.toml
@@ -2,5 +2,5 @@
 [formatter.dprint]
 command = "dprint"
 excludes = []
-includes = [".*"]
+includes = ["*"]
 options = ["fmt"]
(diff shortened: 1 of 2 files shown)


mputz86:
Apparently it is not really correct, but probably best to anyway include only the files for which a plugin was added. So this works:
```
        dprint = {
          enable = true;
          includes = [
            "*.md"
            "*.toml"
            "*.yml"
            "*.yaml"
          ];
          settings.plugins = pkgs.dprint-plugins.getPluginList (plugins:
            with plugins; [
              dprint-plugin-toml
              dprint-plugin-markdown
              g-plane-pretty_yaml
            ]);
        };

```

PR #455, opened 2025-12-29
chore(biome): support 2.3.6 schema

cor:
Adds support for 2.3.6's schema (current `biome` in nixpkgs). I was unsure on if we should prefer not breaking backwards compatibility here or prefer defaulting to a newer schema. Feel free to adjust the PR as needed

### DIFF
--- a/programs/biome.nix
+++ b/programs/biome.nix
@@ -14,10 +14,18 @@ let
 
   cfg = config.programs.biome;
   opts = options.programs.biome;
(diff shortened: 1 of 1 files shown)


cor:
For this to be effective, `pkgs` should also be bumped. IMO we ideally somehow obtain version from the provided `package` to circumvent this.

eveeifyeve:
I have upstreamed here: https://github.com/NixOS/nixpkgs/pull/476453

PR #459, opened 2026-01-05
fix(xmllint): avoid always marking as changed

### LINKED ISSUE #458: xmllint always marks all files as changed
**Describe the bug**

Any time any file is checked for formatting issues by `xmllint`, it is considered to be changed, so `fail-on-change` causes `nix fmt` to always fail.

**To Reproduce**

1. Set `programs.xmllint.enable = true;`
2. Set `settings.fail-on-change = true;`
3. Add a formatted XML file
4. Run `touch *.xml`
5. Run `nix fmt`

At this point, you'll get a "ERRO file has changed" error message and exit code 1.

**Expected behavior**

If the XML files are already formatted, nothing should be printed and the exit code should be zero.

**System information**

 - system: `"x86_64-linux"`
 - host os: `Linux 6.12.63, NixOS, 26.05 (Yarara), 26.05.20251228.c0b0e0f`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.31.2+1`
 - channels(root): `"home-manager, nixos, nixos-hardware, nixos-stable-24.05"`
 - nixpkgs: `/nix/store/xzl5dhddgjxp2p18h5qpkhvlkp8128j1-source`

**Additional context**

The following hack can be used to work around this:

1. Disable `fail-on-change`
2. Manipulate the treefmt cache to think XML files don't need changes by running `nix fmt`
3. Re-enable `fail-on-change`

Subsequent runs will not treat any XML files as changed unless `touch`ed or otherwise changed.

---

It looks like `xmllint` changes the time stamp of every single file, presumably by unconditionally overwriting each of them. This should be avoided - the time stamp should only change if the file was actually changed.

vidhanio:
closes #458

### DIFF
--- a/programs/xmllint.nix
+++ b/programs/xmllint.nix
@@ -34,13 +34,13 @@ in
         name = "xmllint-wrapper";
         text = ''
           temp=$(mktemp)
+          trap 'rm "$temp"' EXIT
           for file in "$@"; do
             ${lib.getExe' cfg.package "xmllint"} --format "$file" --output "$temp"
             if ! cmp -s "$file" "$temp"; then
               cp "$temp" "$file"
             fi
           done
-          rm temp
         '';
       };
     };


vidhanio:
agreed, but xmllint isn't intended to be used as an in-place formatter regardless, so i assume it would be out of scope for them to change.

zimbatm:
thanks!

PR #460, opened 2026-01-05
feat: add `oxfmt` 

jnsgruk:
This PR adds support for [`oxfmt`](https://oxc.rs/docs/guide/usage/formatter).

In order to use `oxfmt`, I had to update the `flake.lock` file, since `oxfmt` was added relatively recently. As a result, I then got evaluation errors for `purs-tidy` and `typstfmt` which have since been removed from `nixpkgs`.

I've taken a best guess at the right way to handle this and tried to keep it in separate commits, but very happy to take guidance on the right approach since this is my first contribution here.


zimbatm:
LGTM, thanks!

PR #461, opened 2026-01-07
fix: remove `*.toml` pattern from `oxfmt`

jnsgruk:
The docs imply that `oxfmt` will format TOML files, but when I run it on a larger project that contains them, it errors out.

This commit removes '*.toml' from the file type list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ functions.
 <!-- `> bash ./supported-programs.sh` -->
 
 <!-- BEGIN mdsh -->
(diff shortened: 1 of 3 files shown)


jnsgruk:
Hmm, I seem to be in a weird state with mdsh, I can't work out what it's doing and each time I correct it, I get a new answer.

jnsgruk:
Okay, fixed it :-)

zimbatm:
thanks! :)

PR #464, opened 2026-01-19
feat(projectRootFile): allow native root detection

a-kenji:
`treefmt-nix` always passes --tree-root-file to treefmt,
which prevents `treefmt` from using its native tree root
detection.

When projectRootFile is set to null, omit `--tree-root-file`
entirely and let `treefmt` detect the tree root itself.

I chose `null` here to be backwards compatible.

Unresolved: I am not sure If I should unset `PRJ_ROOT` here.

### DIFF
--- a/module-options.nix
+++ b/module-options.nix
@@ -143,9 +143,10 @@ in
       description = ''
         File to look for to determine the root of the project in the
         build.wrapper.
+        Set to null to let treefmt use its native detection.
       '';
       default = ".git/config";
-      type = types.str;
+      type = types.nullOr types.str;
     };
 
     enableDefaultExcludes = mkOption {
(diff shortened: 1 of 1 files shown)


jfly (commented):
This seems reasonable, but could you share your use case?

> Unresolved: I am not sure If I should unset PRJ_ROOT here.

This is part of treefmt's native root detection, so I would not unset it. (Related: https://github.com/NixOS/nix/pull/13143)

a-kenji:
My use case is in some projects good marker files might differ and I would prefer to just let `treefmt` be able to handle it and share my configurations around. It's not extremely necessary, but I thought If `treefmt` has the option, why not expose it?

PR #465, opened 2026-01-20
fix: use overrideAttrs to override wrapper.meta

phanirithvij:
I've discovered that overriding the `meta` in the previous way `(x // { meta = config.package.meta // x.meta; })` results in the same derivation, but the `meta` attribute is not retained after it gets passed to another mkDerivation.

```nix
# file: meta-issue.nix
let
  pkgs = import <nixpkgs> { };
  treefmtCfg = import ./treefmt.nix { inherit pkgs; };
  self = ((import ./.).evalModule pkgs treefmtCfg).config.build;
in
{
  config = self;
  inherit (self) wrapper;
  shellDrv = (builtins.elemAt self.devShell.nativeBuildInputs 0);
}
```

`nix repl -f meta-issue.nix`

```nix
nix-repl> wrapper.meta.description
"One CLI to format the code tree"

nix-repl> shellDrv.meta.description
evaluation warning: 'hostPlatform' has been renamed to/replaced by 'stdenv.hostPlatform'
error:
       … while evaluating the attribute 'meta.description'
         at /nix/store/jdiahp8bm5pkl7j8hpim57zmnj6vlcpv-source/pkgs/stdenv/generic/make-derivation.nix:928:17:
          927|         inherit passthru overrideAttrs;
          928|         inherit meta;
             |                 ^
          929|       }

       error: attribute 'description' missing
       at «string»:1:1:
            1| shellDrv.meta.description
             | ^
```

### DIFF
--- a/programs/genemichaels.nix
+++ b/programs/genemichaels.nix
@@ -124,7 +124,7 @@ in
     settingsFile = mkOption {
       description = "The configuration file used by `genemichaels`.";
       type = types.path;
-      example = lib.literalExpression ''./.genemichaels.json'';
+      example = lib.literalExpression "./.genemichaels.json";
       default = configFormat.generate ".genemichaels.json" cfg.settings;
       defaultText = lib.literalMD "Generated JSON file from `${showOptionParent options.programs.genemichaels.settings.max_width 1}`";
     };
--- a/treefmt.nix
+++ b/treefmt.nix
@@ -4,10 +4,10 @@
   settings.excludes = [ "*.toml" ];
 
   programs.deadnix.enable = true;
-  programs.deno.enable = pkgs.hostPlatform.system != "riscv64-linux";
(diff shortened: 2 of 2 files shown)


module-options.nix
```diff
@@ -267,7 +267,9 @@ in
                 '';
             x = pkgs.writeShellScriptBin "treefmt" code;
           in
-          (x // { meta = config.package.meta // x.meta; });
+          x.overrideAttrs (prev: {
+            meta = config.package.meta // prev.meta;
+          });
```
phanirithvij: There's also [lib.meta.addMetaAttrs](https://noogle.dev/f/lib/meta/addMetaAttrs)

But I think we can just use overrideAttrs directly without an abstraction.

zimbatm:
thanks!

PR #466, opened 2026-01-23
feat(tombi): Add tombi formatter

mosauter:
Since the `taplo`-project (see [issue](https://github.com/tamasfe/taplo/issues/715)) has been more or less abandoned there is a better formatter for TOML in [tombi](https://tombi-toml.github.io/tombi).

I don't think any settings for this make sense, as it doesn't have an option to take an explicit config-file and one would need to create a config file at the correct locations but for the moment I'm also happy without it.
But I'm open for suggestions.

(I'm not affiliated in any way, I just like formatting and treefmt-nix, just for clarification sake)


zimbatm (commented):
ok, thanks for the submission

mosauter:
@zimbatm Somehow the pipeline checks seem to got stuck. I rebased it onto the new version but you need to reenable the merge.

musjj:
Hey, any blockers here?

vidhanio:
this seems to be complete @zimbatm, is there any reason that this is blocked on merge?

programs/tombi.nix
```diff
@@ -0,0 +1,12 @@
+{ mkFormatterModule, ... }:
+{
+  meta.maintainers = [ ];
+
+  imports = [
+    (mkFormatterModule {
+      name = "tombi";
+      args = [ "format" ];
```
HeitorAugustoLN, line 8: `--offline` is needed by default to avoid failures on flake checks due to networking restrictions on nix sandboxing

PR #468, opened 2026-01-27
detekt: init

raphiz:
Add support for [detekt](https://detekt.dev/)

It sadly has a weird syntax for the input files (comma separated) - but I was able to fix that with a wrapper


zimbatm:
thanks!

PR #469, opened 2026-01-28
rumdl: init rumdl-check and rumdl-format

delafthi:
Add [rumdl](https://rumdl.dev/) a high-performance Markdown linter and formatter, written in Rust.


senekor:
I'm not sure adding two separate formatters makes sense. `rumdl fmt` is basically an alias for `rumdl check --fix`, except that `rumdl fmt` always exits with code 0, whereas `rumdl check --fix` exits with code 1 if there were rule-violations that cannot be auto-fixed.

delafthi:
I removed the `--fix` flag from `rumdl format` since it has no effect on the command's behavior.

Regarding having two separate commands (format and check): The separation makes kind of sense because `rumdl check` can fail when encountering unsolvable linting issues, whereas I would suspect a formatter not to fail.

However, the problem arises when both commands are enabled. They essentially perform the same operations and only differ in their return values. This seems ugly.

What are your thoughts on this?

senekor:
> However, the problem arises when both commands are enabled. They essentially perform the same operations and only differ in their return values. This seems ugly.
> 
> What are your thoughts on this?

That seems fine. People who know about this can easily avoid it by only activating one option. People who unknowingly activate both will run the same formatting twice unnecessarily. But rumdl is very fast, so it doesn't hurt much.

PR #471, opened 2026-01-29
build.wrapper: optional attribute access

hsjobeki:
For example programs.mypy doesn't provide finalPackage


phanirithvij (approved):

phanirithvij:
Would appreciate if this can be merged @Mic92, @zimbatm.

Unblocks the use of mypy.

PR #501, opened 2026-04-21
json-sort: init

drupol:
JSON-Sort: https://crates.io/crates/json-sort

A CLI tool that sorts JSON object keys to make comparisons and diffs easier, without reordering arrays.

### DIFF
--- a/programs/json-sort.nix
+++ b/programs/json-sort.nix
@@ -1,0 +1,21 @@
+{
+  mkFormatterModule,
+  ...
+}:
+{
+  meta = {
+    maintainers = [
+      "drupol"
+    ];
+  };
+
+  imports = [
+    (mkFormatterModule {
+      name = "json-sort";
+      includes = [
+        "*.json"
+      ];
+      args = [ "--fix" ];
+    })
+  ];
+}
--- a/README.md
+++ b/README.md
@@ -280,6 +280,7 @@ functions.
 * [hujsonfmt](programs/hujsonfmt.nix)
 * [isort](programs/isort.nix)
 * [json-sort-cli](programs/json-sort-cli.nix)
+* [json-sort](programs/json-sort.nix)
 * [jsonfmt](programs/jsonfmt.nix)
 * [jsonnet-lint](programs/jsonnet-lint.nix)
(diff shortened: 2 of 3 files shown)


drupol:
Friendly ping @brianmcgee

programs/json-sort.nix
```diff
@@ -0,0 +1,21 @@
+{
+  mkFormatterModule,
+  ...
+}:
+{
+  meta = {
+    # Example contains store paths
+    skipExample = true;
```
jfly: Does the example *actually* contain store paths? Is it possible you copied this from somewhere?

jfly (commented):
LGTM, just one bit of confusion inline.

programs/json-sort.nix
drupol: I actually don't really understand correctly what's the purpose of this attribute. I most probably copied it from somewhere. Anyway, I just updated the PR and removed it.

programs/json-sort.nix
```diff
@@ -0,0 +1,19 @@
+{
+  mkFormatterModule,
+  ...
+}:
+{
+  meta = {
+    maintainers = [ ];
```
jfly: Mind adding yourself as a maintainer @drupol?

drupol: Sure, why not. Going to fix that.

programs/json-sort.nix
jfly: I think you'll need to regenerate the examples now. Note how CI is now failing:

> The generated ./examples folder is out of sync
> Run ./examples.sh to fix the issue

I think this is why store paths in examples are problematic: they probably change with every nixpkgs bump, and it's just not worth the churn. (Very few of our formatters do this).

drupol:
@jfly Anything else I can do here ?

jfly (approved):
Sorry, I hadn't seen that you updated the PR. LGTM!

drupol:
@zimbatm @jfly Is there anything left in here ? I really wish to see this land :)

jfly:
Still LGTM, but I don't have the commit bit here. @brianmcgee?

PR #503, opened 2026-05-07
feat: add panache-format

jolars:
Adds the Panache formatter (<https://panache.bz>), a formatter for markdown, quarto, and r markdown.


jolars:
CI is failing because pkgs.panache isn't yet in the nixpkgs revision pinned in flake.lock. Panache landed in nixpkgs after 4533d929... (2026-02-03); current nixpkgs-unstable HEAD (68a8af9..., 2026-05-07) has panache 2.41.1 and resolves the eval error.

I haven't included a flake.lock bump in this PR because bumping to current nixpkgs-unstable surfaces an unrelated ols/odinfmt build break (the ols derivation in nixpkgs is out of sync with the newer odin stdlib, filepath.dir arity change). Happy to bump it here if you'd prefer, or to wait for ols upstream to catch up. Whatever's lowest-friction for you?

jolars:
I've now also included an update of the nixpkgs pin in flake.nix, since the ols derivation problem no longer seems to be an issue.

PR #504, opened 2026-05-08
pyright: init

otavio:
## Summary

- Add `programs.pyright` module for the Pyright Python static type checker.
- Mirrors the mypy module's per-directory invocation pattern, since pyright also needs whole-module context to resolve imports rather than running per changed file.
- Drops mypy's `extraPythonPackages`/`extraPythonPaths` because pyright does not consult `PYTHONPATH` — the option description steers users to `extraPaths` in `pyrightconfig.json` / `[tool.pyright]` instead.

## Test plan

- [x] `nix build .#checks.x86_64-linux.formatter-pyright` succeeds and produces a valid treefmt config.
- [x] `nix build .#checks.x86_64-linux.examples` succeeds (pyright is `skipExample`, since the wrapper bakes store paths).
- [x] Multi-directory config evaluated end-to-end: includes resolve to `<dir>/<module>/*.py` and `*.pyi`, and the wrapper invokes `pyright <options> <modules>` from the configured directory.

### DIFF
--- a/programs/pyright.nix
+++ b/programs/pyright.nix
@@ -1,0 +1,115 @@
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  escapePath = lib.replaceStrings [ "/" "." ] [ "-" "" ];
+in
+{
+  meta.maintainers = [ "otavio" ];
+  # Example contains store paths
+  meta.skipExample = true;
+
+  options.programs.pyright = {
+    enable = lib.mkEnableOption "pyright";
+    package = lib.mkPackageOption pkgs "pyright" { };
+    directories = lib.mkOption {
+      description = ''
+        Directories to run pyright in.
+
+        Pyright auto-discovers `pyrightconfig.json` or `pyproject.toml`
+        (`[tool.pyright]`) starting from the directory it is invoked in.
+        Use `extraPaths` in that config to add module search paths —
+        pyright does not read the `PYTHONPATH` environment variable.
+      '';
+      type = lib.types.attrsOf (
+        lib.types.submodule (
+          { name, ... }:
+          {
+            options = {
+              directory = lib.mkOption {
+                type = lib.types.str;
+                default = name;
+                description = ''
+                  Directory to run pyright in, relative to the project root.
+                  The empty string means the project root itself — no `cd`
+                  is performed.
+                '';
+              };
+              options = lib.mkOption {
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
+                example = [ "--warnings" ];
+                description = "Options to pass to pyright";
+              };
+              modules = lib.mkOption {
+                type = lib.types.listOf lib.types.str;
+                default = [ "" ];
+                example = [
+                  "mymodule"
+                  "tests"
+                ];
+                description = "Modules to check (passed as positional file/directory arguments)";
+              };
+            };
+          }
+        )
+      );
+      default = {
+        "" = { };
+      };
+      example = {
+        "" = {
+          modules = [
+            "mymodule"
+            "tests"
+          ];
+        };
+        "subdir" = { };
+      };
+    };
+  };
+
+  config = lib.mkIf config.programs.pyright.enable {
+    settings.formatter = lib.mapAttrs' (
+      name: cfg:
+      let
+        nonEmptyModules = lib.filter (m: m != "") cfg.modules;
+        cdLine = lib.optionalString (cfg.directory != "") "cd ${lib.escapeShellArg cfg.directory}";
+        args = lib.escapeShellArgs (cfg.options ++ nonEmptyModules);
+      in
+      lib.nameValuePair "pyright${lib.optionalString (name != "") "-${escapePath name}"}" {
+        command = pkgs.bash;
+        # Pyright resolves imports from a whole-module context, not per file,
+        # so the wrapper ignores the file path treefmt passes as `$0` and always
+        # invokes pyright on the configured `modules` (or the whole `directory`
+        # when `modules` is empty). With treefmt's batch-size limit this means
+        # pyright may be invoked more than once per directory on large file sets
+        # — each call redundantly type-checks the same modules.
+        #
+        # `treefmt --stdin` is not supported: pyright type-checks the on-disk
+        # modules regardless of what is piped in, so editor integrations should
+        # use pyright's language server directly.
+        options = [
+          "-eucx"
+          ''
+            ${cdLine}
+            exec ${lib.getExe config.programs.pyright.package} ${args}
+          ''
+        ];
+        includes = lib.concatMap (
+          module:
+          let
+            prefix = lib.optional (cfg.directory != "") cfg.directory ++ lib.optional (module != "") module;
+          in
+          [
+            (lib.concatStringsSep "/" (prefix ++ [ "*.py" ]))
+            (lib.concatStringsSep "/" (prefix ++ [ "*.pyi" ]))
+          ]
+        ) cfg.modules;
+      }
+    ) config.programs.pyright.directories;
+  };
+}
--- a/README.md
+++ b/README.md
@@ -315,6 +315,7 @@ functions.
 * [pinact](programs/pinact.nix)
 * [prettier](programs/prettier.nix)
 * [protolint](programs/protolint.nix)
+* [pyright](programs/pyright.nix)
 * [qmlformat](programs/qmlformat.nix)
 * [rstfmt](programs/rstfmt.nix)
 * [rubocop](programs/rubocop.nix)


programs/pyright.nix
```diff
@@ -0,0 +1,96 @@
...
+      '';
+      type = lib.types.attrsOf (
+        lib.types.submodule (
+          { name, ... }:
+          {
+            options = {
+              directory = lib.mkOption {
+                type = lib.types.str;
+                default = name;
+                description = "Directory to run pyright in";
```
jfly: Could you update this description to explain what `""` means? Empty string is not normally something interpreted as a directory in Unix.

programs/pyright.nix
```diff
@@ -0,0 +1,96 @@
...
+  config = lib.mkIf config.programs.pyright.enable {
+    settings.formatter = lib.mapAttrs' (
+      name: cfg:
+      lib.nameValuePair "pyright-${escapePath name}" {
+        command = pkgs.bash;
+        options = [
+          "-eucx"
+          ''
+            ${lib.optionalString (cfg.directory != "") ''cd "${cfg.directory}"''}
+            ${lib.getExe config.programs.pyright.package} ${lib.escapeShellArgs cfg.options} ${lib.escapeShellArgs cfg.modules}
```
jfly: nit: avoid an unnecessary intermediate bash process with `exec`

programs/pyright.nix
```diff
@@ -0,0 +1,96 @@
...
+          ''
+        ];
+        includes = lib.concatMap (
```
jfly, line 102: I don't understand this `includes` directive. If for a given directory it matches 2 `*.py` files, won't treefmt invoke pyright twice for the same directory?

programs/pyright.nix
```diff
@@ -0,0 +1,96 @@
...
+  pkgs,
+  config,
+  ...
+}:
+let
+  escapePath = lib.replaceStrings [ "/" "." ] [ "-" "" ];
+in
+{
+  meta.maintainers = [ "otavio" ];
+  # Example contains store paths
```
jfly, line 12: Just curious: where are the problematic store paths?

otavio:
Self-review pass on the `/review` output, with each item resolved. Fixes are folded into the existing `pyright: init` commit (amended).

**1. Default `modules = [ "" ]` passes empty positional arg to pyright** — *fixed*. `cfg.modules` is now filtered with `lib.filter (m: m != "")` before `escapeShellArgs`, and the positional-arg fragment is omitted entirely when the filtered list is empty. The default invocation is now `pyright` (no trailing `''`), so pyright auto-discovers from cwd / `pyrightconfig.json` cleanly. This intentionally diverges from `programs/mypy.nix`, which has the same pattern and arguably the same latent issue.

**2. `*.py` only matches top-level files** — *withdrawn*. I was wrong: treefmt globs are recursive by default, and every other Python formatter in the repo (`black`, `ruff-format`, `ruff-check`, `yapf`, `isort`) uses bare `*.py`/`*.pyi` the same way. No change needed.

**3. Trailing dash in formatter name (`pyright-`)** — *fixed*. The default key now produces `[formatter.pyright]`; non-empty keys still produce `pyright-<escaped-key>` (e.g. `pyright-src-foo`). Implemented with `lib.optionalString (name != "") "-${escapePath name}"`.

**4. `cd "${cfg.directory}"` not shell-escaped** — *fixed*. Now uses `lib.escapeShellArg cfg.directory`, hardening against directories containing quotes, `$`, or whitespace.

**5. Test plan describes config-eval as "end-to-end"** — *clarified, no code change*. The `formatter-pyright` check generates a TOML config and validates evaluation; this matches every other formatter in the repo and is the established convention. Adding runtime invocation tests would set a precedent for ~140 formatters and is out of scope here.

**6. `meta.maintainers = [ "otavio" ]`** — acknowledged, no action.

**Bonus follow-up (not in this PR):** `programs/mypy.nix` only emits `*.py` in `includes`, missing `*.pyi`. Worth a separate small PR — mypy reads stub files too.

**Verification of the amended commit:**
- `nix build .#checks.x86_64-linux.formatter-pyright` — passes; emitted TOML shows `[formatter.pyright]` with `includes = ["*.py", "*.pyi"]` and no empty positional arg.
- `nix build .#checks.x86_64-linux.examples` — passes (pyright remains `skipExample`).
- `nix build .#checks.x86_64-linux.self-formatting` — passes after applying nixfmt's split of the long interpolation line.
- Multi-directory smoke (`directories."src/foo" = { modules = [ "mymod" ]; };`) evaluates to formatter name `pyright-src-foo`, `cd src/foo`, and `includes = ["src/foo/mymod/*.py", "src/foo/mymod/*.pyi"]`.

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  escapePath = lib.replaceStrings [ "/" "." ] [ "-" "" ];
+in
+{
+  meta.maintainers = [ "otavio" ];
```
otavio, line 11: **Original concern #6: `meta.maintainers = [ "otavio" ]` improves over mypy's empty list** — acknowledged, no action needed. Will resolve.

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+  # Example contains store paths
+  meta.skipExample = true;
```
otavio, line 13: **Original concern #5: test plan described config-eval as "end-to-end"** — clarification only, no code change. The `formatter-pyright` CI check generates a TOML config and validates evaluation, which is the established convention for every formatter in this repo (`meta.skipExample = true` correctly excludes from the examples runtime check, since the wrapper bakes store paths). Adding real runtime invocation tests would set a precedent across ~140 formatters and is out of scope here. Leaving open since it's a doc/clarification rather than a fix.

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+        "subdir" = { };
+      };
+    };
+  };
+
+  config = lib.mkIf config.programs.pyright.enable {
+    settings.formatter = lib.mapAttrs' (
+      name: cfg:
+      let
+        nonEmptyModules = lib.filter (m: m != "") cfg.modules;
```
otavio, line 79: **Original concern #1: default `modules = [ "" ]` passed an empty positional arg `''` to pyright** — addressed. `nonEmptyModules` filters empties out, and `modulesArgs` on line 77 only emits the positional-args fragment when the filtered list is non-empty. The default invocation is now a clean `pyright` with no trailing empty arg, so pyright auto-discovers from cwd / `pyrightconfig.json` cleanly. Intentionally diverges from `programs/mypy.nix`, which has the same latent issue.

otavio (commented):
Self-review pass mapping each original change request to its inline location in the amended diff. See per-line comments below.

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+        cdLine = lib.optionalString (cfg.directory != "") "cd ${lib.escapeShellArg cfg.directory}";
```
otavio, line 80: **Original concern #4: `cd "${cfg.directory}"` was not shell-escaped** — addressed. Now uses `lib.escapeShellArg`, hardening against directories that contain quotes, `$`, or whitespace.

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+        modulesArgs = lib.optionalString (nonEmptyModules != [ ]) " ${lib.escapeShellArgs nonEmptyModules}";
+      in
+      lib.nameValuePair "pyright${lib.optionalString (name != "") "-${escapePath name}"}" {
```
otavio, line 83: **Original concern #3: trailing `pyright-` formatter name when the key is `""`** — addressed. `lib.optionalString (name != "") "-${escapePath name}"` drops the dash for the default key, so the generated TOML is `[formatter.pyright]` instead of `[formatter.pyright-]`. Non-empty keys still produce `pyright-<escaped-key>` (e.g. `pyright-src-foo`).

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+        command = pkgs.bash;
+        options = [
+          "-eucx"
+          ''
+            ${cdLine}
+            exec ${lib.getExe config.programs.pyright.package} ${lib.escapeShellArgs cfg.options}${modulesArgs}
```
otavio: Follow-up from a `/simplify` pass: pre-bound `cdLine` and `modulesArgs` in the `let` for readability (the bash heredoc now reads top-to-bottom without nested `${lib.optionalString …}` interpolation), and added `exec` so bash hands off to pyright instead of staying alive holding fds during type-checking. Small per-invocation runtime win.

programs/pyright.nix
```diff
@@ -0,0 +1,101 @@
...
+          ''
+        ];
+        includes = lib.concatMap (
+          module:
+          let
+            prefix = lib.optional (cfg.directory != "") cfg.directory ++ lib.optional (module != "") module;
+          in
+          [
+            (lib.concatStringsSep "/" (prefix ++ [ "*.py" ]))
```
otavio, line 108: **Original concern #2: `*.py` would only match top-level files** — **withdrawn (reviewer error).** treefmt globs are recursive by default, and every other Python formatter in the repo (`black.nix:8-11`, `ruff-format.nix:26-29`, `ruff-check.nix:37-40`, `yapf.nix:8-10`, `isort.nix:16-19`) uses bare `*.py`/`*.pyi` exactly the same way. No change needed.

programs/pyright.nix
otavio: Updated in 40d2d4e — the `directory` option now explicitly documents that the empty string means the project root (no `cd` is performed). The same convention applies to the default attrset key `""` of the parent `directories`. Resolving.

programs/pyright.nix
otavio: Caught this independently in a `/simplify` pass and added `exec` (along with pre-binding `cdLine` / `modulesArgs` in the `let` for readability) — see line 86 of the amended commit. Resolving.

programs/pyright.nix
otavio: Good question — no. treefmt invokes the formatter once per batch with the matched files appended as positional args (not once per matched glob pattern). The wrapper script deliberately ignores those positional args: `exec pyright <opts> <configured-modules>` never references `$@`, because pyright needs whole-module context to resolve imports rather than a per-file invocation. So matching both `*.py` and `*.pyi` for the same directory still results in at most one pyright invocation per directory, operating on the configured `modules` (or the whole `directory` when `modules` is empty). Same pattern as `programs/mypy.nix`. Happy to add a code comment documenting this if you'd prefer — leaving open for your call.

programs/pyright.nix
otavio: The bash heredoc embeds `${lib.getExe config.programs.pyright.package}`, which expands to `/nix/store/<hash>-pyright-<version>/bin/pyright` inside the `options` field of the generated TOML. The example-generator in `checks/default.nix` only strips store paths from the `command = ...` line via sed, not from inside `options`, so a generated example would leak a pinned-version pyright store path. `programs/mypy.nix` uses `meta.skipExample = true` for the same reason. Leaving open in case you'd like the in-code comment expanded.

programs/pyright.nix
jfly: Ha! I suppose that works, modulo the fact that treefmt has a max batch size. This definitely needs a comment.

programs/pyright.nix
jfly: Nope, that's fine.

programs/pyright.nix
otavio: Added the comment in 101eda1 — six lines above the `options` heredoc documenting that the wrapper ignores treefmt's appended file args, why (whole-module context), and the batch-size implication you raised. Resolving.

programs/pyright.nix
otavio: Thanks — resolving.

otavio:
Update on top of [the earlier summary](https://github.com/numtide/treefmt-nix/pull/504#issuecomment-4407040804) — all review threads are now resolved. Additional changes folded into the same `pyright: init` commit (current HEAD `101eda1`):

- **`directory` option description** — expanded to document that the empty string (and the default attrset key `""`) means the project root, with no `cd` performed. Addressing @jfly's request for clarification.
- **`exec` before pyright invocation** — independently caught in a `/simplify` pass before seeing @jfly's `nit`. Bash now hands off rather than staying alive holding fds during type-checking.
- **`let`-bound `cdLine` / `modulesArgs`** — pre-binding pulled out of the bash heredoc so the multi-line string reads top-to-bottom without nested `${lib.optionalString …}` interpolation.
- **Wrapper-ignores-args comment** — added a 6-line comment above `options` documenting that pyright is always invoked on the configured `modules` (or the whole `directory`), not on the matched files treefmt appends as positional args, and noting the implication of treefmt's batch-size limit (multiple invocations per directory possible on large file sets, each operating on the same modules). Addressing @jfly's request on the `includes` thread.
- **`meta.skipExample = true` rationale** — explained on the original thread that the embedded `lib.getExe` store path inside `options` is what defeats the example sed-stripper; @jfly accepted, no code change needed.

CI is green (`formatter-pyright`, `examples`, `self-formatting` all pass). Ready for another look.

programs/pyright.nix
```diff
@@ -0,0 +1,111 @@
...
+        lib.types.submodule (
+          { name, ... }:
+          {
+            options = {
+              directory = lib.mkOption {
+                type = lib.types.str;
+                default = name;
+                description = ''
+                  Directory to run pyright in, relative to the project root.
+                  The empty string (also the default attrset key) means the
```
jfly: nit: I'm not sure that "the default attrset key" is a very comprehensible phrase. I'd consider removing it. (Feel free to mark this resolved if you disagree).

programs/pyright.nix
```diff
@@ -0,0 +1,111 @@
...
+        # so the wrapper ignores the matched files treefmt appends as positional
+        # args and always invokes pyright on the configured `modules` (or the
+        # whole `directory` when `modules` is empty). With treefmt's batch-size
+        # limit this means pyright may be invoked more than once per directory
+        # on large file sets — each call still type-checks the same modules.
+        options = [
+          "-eucx"
+          ''
+            ${cdLine}
+            exec ${lib.getExe config.programs.pyright.package} ${lib.escapeShellArgs cfg.options}${modulesArgs}
```
jfly: nit: I think this code would be cleaner with one call to `escapeShellArgs` rather than one for `cfg.options`, and another when building `modulesArgs`.

programs/pyright.nix
```diff
@@ -0,0 +1,111 @@
...
+        modulesArgs = lib.optionalString (nonEmptyModules != [ ]) " ${lib.escapeShellArgs nonEmptyModules}";
+      in
+      lib.nameValuePair "pyright${lib.optionalString (name != "") "-${escapePath name}"}" {
+        command = pkgs.bash;
+        # Pyright resolves imports from a whole-module context, not per file,
+        # so the wrapper ignores the matched files treefmt appends as positional
+        # args and always invokes pyright on the configured `modules` (or the
+        # whole `directory` when `modules` is empty). With treefmt's batch-size
+        # limit this means pyright may be invoked more than once per directory
+        # on large file sets — each call still type-checks the same modules.
```
jfly: Emphasis mine:

> on large file sets — each call **still** type-checks the same modules.

nit: I'm not understanding what the word "still" is doing here. How about "redundantly" instead?

programs/pyright.nix
```diff
@@ -0,0 +1,111 @@
...
+      let
+        nonEmptyModules = lib.filter (m: m != "") cfg.modules;
+        cdLine = lib.optionalString (cfg.directory != "") "cd ${lib.escapeShellArg cfg.directory}";
+        modulesArgs = lib.optionalString (nonEmptyModules != [ ]) " ${lib.escapeShellArgs nonEmptyModules}";
+      in
+      lib.nameValuePair "pyright${lib.optionalString (name != "") "-${escapePath name}"}" {
+        command = pkgs.bash;
+        # Pyright resolves imports from a whole-module context, not per file,
+        # so the wrapper ignores the matched files treefmt appends as positional
+        # args and always invokes pyright on the configured `modules` (or the
```
jfly: > ignores the matched files treefmt appends as positional args 

Does this work with treefmt's `--stdin` mode?

programs/pyright.nix
otavio: Agreed — dropped the parenthetical in 86a4b59. The description now reads: "Directory to run pyright in, relative to the project root. The empty string means the project root itself — no `cd` is performed." Resolving.

programs/pyright.nix
otavio: Done in 86a4b59 — combined into a single `args = lib.escapeShellArgs (cfg.options ++ nonEmptyModules)` binding, dropping the separate `modulesArgs` (and its `lib.optionalString` guard, since `escapeShellArgs []` is already an empty string). Generated TOML is identical. Resolving.

programs/pyright.nix
otavio: Good catch — "redundantly" is more precise. Updated in 86a4b59. Resolving.

programs/pyright.nix
otavio: Verified empirically. Built a small test project with `programs.pyright.enable = true` and a minimal `pyrightconfig.json`, then ran:

```
echo '<fixed code>' | treefmt --stdin foo.py
```

while leaving an obvious type error in the on-disk `foo.py`. Result: pyright was invoked, type-checked the on-disk modules per the config, reported the error against `foo.py:2:12`, and treefmt wrote the *original stdin content* unchanged to stdout while exiting non-zero with `failed to finalise formatting`.

So `--stdin` mode "works" in the sense that it doesn't crash, but the wrapper deliberately ignores stdin (consistent with how it ignores treefmt's appended file args in normal mode) — pyright always type-checks the on-disk modules, never the piped content. This is the same trade-off `programs/mypy.nix` makes: type-checkers fundamentally need whole-module context. The wrapper comment already documents the per-file ignore; `--stdin` is a special case of the same behavior. Happy to extend the comment to call out `--stdin` explicitly if you'd like — leaving open for your call.

jfly: > Happy to extend the comment to call out --stdin explicitly if you'd like — leaving open for your call.

Sure, I suppose a comment is the best we can do for now. Seems like the better choice vs discarding this PR. Did you say you copied this pattern from other formatters (mypy?). Would be nice to add corresponding comments over there as well if you're up for it.

> the wrapper deliberately ignores stdin

This implies incorrectly that the wrapper receives anything via stdin. It doesn't, treefmt creates a temp file and passes that path to the formatter. (Only relevant for the comment you're going to add.)

otavio: Decided to fix this directly rather than just document it. After probing treefmt-go's invocation surface, I confirmed that in `--stdin` mode treefmt writes the buffer to a temp file whose path matches `*/treefmt-stdin-*` and passes it as the bash `$0` argument. The wrapper now detects that path and exits 0 immediately:

```bash
case "$0" in */treefmt-stdin-*) exit 0 ;; esac
```

Why skip rather than try to type-check the stdin content: pyright fundamentally needs whole-module context to resolve imports — type-checking an isolated buffer in isolation would lose all cross-file type info. Skipping is the right semantic for a type-checker-as-formatter when invoked on a single buffer.

Verified end-to-end against a project with a deliberate type error:
- `treefmt foo.py` → pyright runs (387ms), surfaces the error, exits non-zero. ✓
- `treefmt --stdin foo.py` → pyright skipped (3ms), stdin passes through unchanged, exits 0. ✓

In d3613c9. Repo CI (`formatter-pyright`, `examples`, `self-formatting`) all green. Resolving.

jfly: I don't agree with the latest change.

> mode treefmt writes the buffer to a temp file whose path matches */treefmt-stdin-*

Speaking as a treefmt maintainer, this is an implementation detail of treefmt that is *not* ok to depend upon. Furthermore, what if my project has a file that matches that glob?

otavio: Ok, please guide me on how you prefer to handle this situation. There is any example that has this, as you think that would be the best way to handle the situation?

otavio: Fair points on both counts — depending on treefmt's `treefmt-stdin-` prefix is reaching into an implementation detail you'd rightly want freedom to change, and the collision risk against real project files is real.

Reverted the `case` guard in 1747944. Probed every reachable signal from inside a treefmt formatter (env vars, `$0`, `$#`, stdin pipe/tty, cwd) and there is no stable in-band signal that distinguishes `--stdin` mode from a normal per-file invocation today, so I'm dropping the detection attempt entirely rather than substituting another brittle heuristic.

Instead, the `directories` option description now documents the limitation explicitly:

> Pyright always type-checks the configured on-disk modules and does not consume `treefmt --stdin` input. For editor integration, use pyright's language server directly.

This matches how `programs/mypy.nix` behaves today (same trade-off, same lack of detection) and is honest about the constraint. If treefmt would consider exposing a stable signal upstream — e.g. a `TREEFMT_STDIN=1` env var passed to formatters in stdin mode — that would let `programs.pyright` (and mypy, and other type-checkers/linters in similar positions) opt out cleanly. Happy to file an issue if that direction is welcome; otherwise this PR is in its final shape from my side.

jfly: > Ok, please guide me on how you prefer to handle this situation

I already said what I prefer earlier: "Sure, I suppose a comment is the best we can do for now".

> e.g. a TREEFMT_STDIN=1 env var passed to formatters in stdin mode — that would let programs.pyright (and mypy, and other type-checkers/linters in similar positions) opt out cleanly. Happy to file an issue if that direction is welcome

Please see https://github.com/numtide/treefmt/pull/586 if you want to learn more about this.

jfly (approved):

programs/pyright.nix
otavio: Got it — moved the `--stdin` note out of the `directories` option description and into a code comment in the wrapper itself (66f38f7), where it sits alongside the existing comment about pyright operating on whole-module context. The comment also references numtide/treefmt#586 so a future reader knows where the clean opt-out is heading. Thanks for the pointer to the Stdin Specification work.

programs/pyright.nix
```diff
@@ -0,0 +1,117 @@
...
+        # Pyright resolves imports from a whole-module context, not per file,
+        # so the wrapper ignores the file path treefmt passes as `$0` and always
+        # invokes pyright on the configured `modules` (or the whole `directory`
+        # when `modules` is empty). With treefmt's batch-size limit this means
+        # pyright may be invoked more than once per directory on large file sets
+        # — each call redundantly type-checks the same modules.
+        #
+        # `treefmt --stdin` is not supported: pyright type-checks the on-disk
+        # modules regardless of what is piped in, so editor integrations should
+        # use pyright's language server directly. A clean opt-out will become
```
jfly: What does "clean opt-out" mean?

otavio: Fair point — re-reading #586, it adds a path for formatters to opt **into** stdin handling via `stdin-options`, not an opt-out for formatters like pyright that don't fit stdin mode at all. "Clean opt-out" was speculative phrasing that doesn't accurately describe what #586 provides. Dropped the sentence (and the #586 reference) in d39d69d, leaving the comment as just the factual statement: `treefmt --stdin` isn't supported, use the language server for editor integration. Thanks for catching the confusing language.

jfly (approved):

PR #507, opened 2026-05-20
pkl: init

tkoyasak:
Adds https://pkl-lang.org/ to treefmt. Pkl ships `pkl format` (since 0.28); the formatter module wires `pkl format --write` so treefmt rewrites files in place. `includes` covers both `*.pkl` module files and the extension-less `PklProject` project file, which is also Pkl syntax.


programs/pkl.nix
```diff
@@ -0,0 +1,18 @@
+{ mkFormatterModule, ... }:
+{
+  meta.maintainers = [ ];
```
jfly: Would you consider adding yourself as a maintainer, @tkoyasak?

jfly (approved):

programs/pkl.nix
tkoyasak: Sure, done! Added myself as a maintainer in 7ce7e62. Thanks for the review 🙏

PR #509, opened 2026-05-27
feat: Add `extendIgnore` option to `programs.ruff-check`

qhga:
Adds an additional `extendIgnore` option to the `ruff-check` linter to provide aditional rule codes that should be ignored.


qhga:
Closing in favor of #520

PR #514, opened 2026-06-09
default `projectRootFile` to `null`

### LINKED ISSUE #373: Formatter does not work in a git worktree
**Describe the bug**

Running `nix fmt` (set up as recommended in the README) fails with
```
Error: failed to load config: failed to determine tree root: failed to find tree-root based on tree-root-file: could not find [.git/config] in /path/to/your/worktree
````

**To Reproduce**

1. Create a git worktree using `git worktree add`
2. Try to run `nix fmt` in the worktree.

**Expected behavior**

Formats as normal.

**System information**

treefmt-nix version: a05be418a1af1198ca0f63facb13c985db4cb3c5
nixpkgs version: 56b033fe4f9da755b1872466f24b32df7cfc229e

`nix-shell -p nix-info --run "nix-info -m"`:
```
 - system: `"aarch64-linux"`
 - host os: `Linux 6.16.0-rc3, NixOS, 25.11 (Xantusia), 25.11.20250615.cbfb263`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.91.1
System type: aarch64-linux
Additional system types: 
Features: gc, signed-caches
System configuration file: /etc/nix/nix.conf
User configuration files: /home/user/.config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/user/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/user/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/user/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/g9clkwljsncp6fadzc105ay78b7l6mg2-lix-2.91.1/share`
 - nixpkgs: `/nix/store/nki040zkkasz5mnd7c6jv0gn3sipmc30-source`
```

### PR DESCRIPTION
**Pull request by KiaraGrouwstra**
Lets `treefmt` figure out the project root to fix root detection in git worktrees, implementing https://github.com/numtide/treefmt-nix/issues/373#issuecomment-3017042316.

Closes #373.

### DIFF
--- a/module-options.nix
+++ b/module-options.nix
@@ -145,7 +145,7 @@ in
         build.wrapper.
         Set to null to let treefmt use its native detection.
       '';
-      default = ".git/config";
+      default = null;
       type = types.nullOr types.str;
     };
 
--- a/module-options.nix
+++ b/module-options.nix
@@ -143,9 +143,9 @@ in
       description = ''
         File to look for to determine the root of the project in the
         build.wrapper.
-        Set to null to let treefmt use its native detection.
+        Leave set to null to let treefmt use its native detection.
       '';
-      default = ".git/config";
+      default = null;
       type = types.nullOr types.str;
     };
 

### CODE REVIEW

File: module-options.nix [hunk 1]
```diff
@@ -145,7 +145,7 @@ in
         build.wrapper.
         Set to null to let treefmt use its native detection.
```
MattSturgeon:
```suggestion
         Leave set to null to let treefmt use its native detection.
```

Not sure about my suggested wording, but I think we should re-phrase the above advice if it will be the default behaviour.

jfly (approved):

MattSturgeon (approved):

jfly:
Let's figure out https://github.com/numtide/treefmt-nix/issues/540 before landing this.

MattSturgeon:
It may be worth noting that this adds a subtle dependency on `git` (or `jj` or another tool treefmt supports) being installed on the PATH and the project being tracked by version control.

This shouldn't be an issue for most users, as anyone affected can just define `projectRootFile`.

If there is a good way to warn affected users without spamming unaffected users, it'd be nice to see it. Other than wrapping the `treefmt` binary to detect the new failure mode and print an additional instruction to use `projectRootFile`, I can't think of anything. Therefore, I still approve of this PR as-is.

jfly:
> and the project being tracked by version control

The previous value (`.git/config`) arguably is more restrictive (it strongly implies you have the git cli, as a source tarball wouldn't provide it, and doesn't support jj).

MattSturgeon:
Yes. I'm just pointing out that the failure mode is subtly different. The runtime requirements change from files in the project directory to executables installed on PATH.

I'd forgotten that the actual default was in `.git`, and I agree that strongly implies a user cloned or created their project using git and likely has git permanently installed.

In theory, it's possible that git was transiently installed or run via `nix run`, or it's possible the user manually created a `.git/config` file to satisfy the default. Both of these are sufficiently niche that they're not worth accommodating!

PR #517, opened 2026-06-30
fix: add `*.heex` pattern to `mix-format`

datafoo:
This pattern is for the Phoenix HEEx template language.


datafoo:
@mergifyio queue

PR #523, opened 2026-07-18
Drop x86_64-darwin from supported systems

### LINKED ISSUE #521: Compatibility with Nixpkgs 26.11: x86_64-darwin dropped from supported platforms
**Заголовок:** Compatibility with Nixpkgs 26.11: `x86_64-darwin` dropped from supported platforms

Nixpkgs 26.11 has dropped support for `x86_64-darwin` ([PR #493096](https://github.com/NixOS/nixpkgs/pull/493096)). The `eachSystem` list in `treefmt-nix`'s `flake.nix` still includes `"x86_64-darwin"`, which causes evaluation to fail when `treefmt-nix` is used as a dependency in flakes that track `nixpkgs-unstable` (post-26.11):

```
error: Nixpkgs 26.11 has dropped support for x86_64-darwin.
```

The `formatter.<system>` and `checks.treefmt` attributes are unconditionally defined for all four systems, including `x86_64-darwin`. When Nixpkgs can no longer be instantiated for that platform, the entire flake evaluation fails.

### Proposed solutions

1. **Drop `x86_64-darwin` from `eachSystem`** — simplest fix, aligns with upstream direction. Users on Intel Macs can pin to an older revision.

2. **Make `eachSystem` overridable** — introduce a `systems` input (like `flake-utils` does) so consumers can control which systems are evaluated without forking:

   ```nix
   inputs.systems.url = "github:nix-systems/default";
   ```

3. **Drop `x86_64-darwin` by default, add a flag to re-enable it** — remove `x86_64-darwin` from `eachSystem` and add an opt-in mechanism (e.g. an input override or a top-level option) for the few users who still need it on Intel Macs.

Personally I'd prefer solution 1 for its simplicity, but I'd like to hear what the maintainers think. @zimbatm @Mic92 @srid @brianmcgee

xilec:
Nixpkgs 26.11 dropped support for `x86_64-darwin` (NixOS/nixpkgs#493096), which makes this flake fail to evaluate when used as a dependency on `nixpkgs-unstable` (post-26.11):

    error: Nixpkgs 26.11 has dropped support for x86_64-darwin.

Fixes #521

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,6 @@
       eachSystem = lib.genAttrs [
         "aarch64-darwin"
         "aarch64-linux"
-        "x86_64-darwin"
         "x86_64-linux"
       ];
     in


xilec:
@Mic92 Could you please take a look when you have a moment? Thanks!

PR #532, opened 2026-08-14
Add support for stdin mode, update prettier to use it

jfly:
I suggest reading the commits independently.

tl;dr, this fixes https://github.com/numtide/treefmt-nix/issues/530


jfly:
@brianmcgee, now that treefmt 2.6.0 has been released, could I get a review here?

default.nix
```diff
@@ -13,6 +13,7 @@ let
       package ? name,
       mainProgram ? null,
       args ? [ ],
+      stdin-args ? null,
```
lunagl, line 16: Should probably be `stdinArgs` to keep with the naming scheme used everywhere else

jfly: Oops, fixed. Good call.

lunagl (changes requested):
Can you add `stdin-options` below `module-options.nix:84` with the docs taken from [here](https://treefmt.com/latest/getting-started/configure/#stdin-options)?

> I didn't copy the existing `_type or null == "order"` stuff is. AFAICT,
it's not used anywhere for `options`, but perhaps I'm reading it wrong.

This makes the args given to `mkFormatterModule` default to `mkBefore` if not explicitly given an ordering. I don't know if it's entirely necessary, but would make sense to keep `args` and `stdinArgs` consistent in this regard.
You're right that we don't pass an explicit ordering anywhere at the moment, but _if_ we want to default to `mkBefore`, this check should remain, as the behavior would be confusing without it.

jfly:
> Can you add stdin-options below module-options.nix:84 with the docs taken from [here](https://treefmt.com/latest/getting-started/configure/#stdin-options)?

You nerd sniped me here. It is possible, but it's a little tricky as the nix module system doesn't allow for representing omittable values (as a workaround, I allow for null and then strip out nulls): https://github.com/NixOS/nixpkgs/issues/563975.

> You're right that we don't pass an explicit ordering anywhere at the moment, but if we want to default to mkBefore, this check should remain, as the behavior would be confusing without it.

I agree, we should keep things consistent. I've opted to change this code to match the other code, but I'm not usually a fan of dead code, so I'd gladly delete the other code to match instead. Let me know what you think.

lunagl:
> You nerd sniped me here. It is possible, but it's a little tricky as the nix module system doesn't allow for representing omittable values (as a workaround, I allow for null and then strip out nulls): https://github.com/NixOS/nixpkgs/issues/563975.

Right, I didn't even consider that. The solution with filtering out nulls seems good, thank you!

> I agree, we should keep things consistent. I've opted to change this code to match the other code, but I'm not usually a fan of dead code, so I'd gladly delete the other code to match instead. Let me know what you think.

Removing it would change the behavior of all modules where additional options are set in `config` (grep for `options =` in programs/ for a list), because the `args` passed to `mkFormatterModule` wouldn't be guaranteed to be ordered before those other options.

lunagl (approved):

jfly:
[CI failure](https://buildbot.numtide.com/#/builders/314/builds/700) seems to be unrelated:

> ```
> copying path '/nix/store/b7155j8rnmp6c6y1wnck9v24y1jhh7cg-treefmt' to 'grpc://farm.numtide.com:50051'...
> error: gRPC AddMultipleToStore on 'farm.numtide.com:50051' failed: error: path '/nix/store/0qdx7ah7b1dwyxa03fblhcdjnpi29q25-nixfmt-1.2.0' is not valid
> error: Cannot build '/nix/store/gfgqbxj4w4swar7aw7gm17yjcpmp1xhl-treefmt-check.drv'.
>        Reason: builder failed with exit code 1.
>        Output paths:
>          /nix/store/z86sksh4blz8fq34wcm8fypij2a94zzb-treefmt-check
> ```

PR #536, opened 2026-09-10
feat(typos): add autoFix option

### LINKED ISSUE #535: typos: Add an `autoFix` option to opt in/out of automatic typo correction
**Is your feature request related to a problem? Please describe.**

The default for formatting with `typos` is that typos are automatically "corrected."

**Describe the solution you'd like**

It should be possible to opt out of this behaviour, since there are many instances where `typos` actually gets it wrong and makes destructive changes, which isn't true of most other formatters.

**Describe alternatives you've considered**

I can override the flags passed to `typos` using:

```nix
perSystem = {
  treefmt = {
    settings.formatter.typos.options = lib.mkForce [
      # Override options here...
    ];
  };
};
```

But I'd prefer not to have to use this escape hatch, because then I lose out on all of the built-in `typos` module options.

dnlzro:
Add an `autoFix` option to the `typos` formatter configuration. Default behaviour is unchanged (i.e., `autoFix = true`), but now users can easily opt-out.

Closes #535.

### DIFF
--- a/programs/typos.nix
+++ b/programs/typos.nix
@@ -29,7 +29,7 @@ in
       description = ''
         Write fixes out
 
-        Set to falkse to run typos in check-only mode; useful if you find that
+        Set to false to run typos in check-only mode; useful if you find that
         automatic corrections are often destructive or incorrect.
       '';
     };
--- a/programs/typos.nix
+++ b/programs/typos.nix
@@ -14,8 +14,6 @@ in
     (mkFormatterModule {
       name = "typos";
       args = [
-        "--write-changes"
-
         # Treefmt may pass files otherwise ignored by typos (e.g. files ignored in typos.toml).
         # '--force-exclude' stops typos from acting on any ignored files passed
         "--force-exclude"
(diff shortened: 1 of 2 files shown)


programs/typos.nix
```diff
@@ -25,6 +23,12 @@ in
   ];
 
   options.programs.typos = {
+    autoFix = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "Write fixes out";
```
jfly: Maybe worth a little elaboration why you might want to turn this off? You already gave a good explanation in https://github.com/numtide/treefmt-nix/issues/535.

jfly (approved):
LGTM! One non-blocking suggestion for you.

programs/typos.nix
dnlzro: I appreciate the feedback! The description has been improved. (BTW, the first line is just taken from the `typos --help` option description, consistent with the other options from before this PR.)

programs/typos.nix
```diff
@@ -25,6 +23,17 @@ in
   ];
 
   options.programs.typos = {
+    autoFix = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        Write fixes out
+
+        Set to falkse to run typos in check-only mode; useful if you find that
```
jfly:
```suggestion
        Set to false to run typos in check-only mode; useful if you find that
```

dnlzro:
Ironic 🤦 Should be good to merge now.

jfly (approved):

PR #538, opened 2026-09-17
starlark_fmt: init module

lunagl:
Based on top of #532 for stdin mode support.

First commit updates flake inputs to get treefmt 2.6 and a buck2 with starlark_fmt (was only recently added).


programs/starlark_fmt.nix
```diff
@@ -0,0 +1,147 @@
...
+  meta.maintainers = [ "lunagl" ];
+  # example contains store paths
+  meta.skipExample = true;
+
+  imports = [
+    (mkFormatterModule {
+      name = "starlark_fmt";
+      package = "buck2";
+      mainProgram = "starlark_fmt";
+      # order between mkBefore (500) and default (1000)
```
jfly, line 50: I don't understand what we're doing here with `mkOrder`. This seems like something that would only matter when merging multiple modules, which we're not doing here. Is this enabling some use case I cannot see in this PR?

jfly (commented):
LGTM, one question for you inline.

programs/starlark_fmt.nix
lunagl: the default would be mkBefore, but I want the `--config` arg to be the first thing. the 750 isn't super important, I could've also chosen 1000, but I thought it might be nice to keep these ordered before regular default ordering, in case a user decides to add something I suppose.

jfly: Ok, this is embarrassing: I've been really misreading these modules: I now understand that the `args` we pass to `mkFormatterModule` get merged with the `settings.formatter.${name}.options` we often set in the `config` block below.

Now I understand why you're thinking so much about ordering!

jfly: (I tried to mark this thread resolved, but cannot. Feel free to do so yourself)

